PR #1, opened 2022-04-01
create skeleton of app

Gankra:
This is mostly me setting up my "preferred" machinery for a CLI app (although this is my first time using clap's derive-based machinery, I am quite liking it here).

Some extra stuff like --verbose, --log-file, --output-file are just "standard" things I like a CLI app to have for debugging and making them more flexible to the user.

What actually is implemented:

* minimal `vet init`
* trivial `vet` (does not apply trusted/audits or deltas yet)

A lot of the machinery set up here should make many of the other commands relatively trivial (open toml, push value to a Vec, write toml back).


src/main.rs
```diff
@@ -1,3 +1,829 @@
...
+    /// Yield a diff against the last reviewed version.
+    Diff(DiffArgs),
+    /// Mark `$crate $version` as reviewed with `$message`
+    Certify(CertifyArgs),
+    /// Mark `$crate $version` as unacceptable with `$message`
+    Forbid(ForbidArgs),
+    /// Suggest some low-hanging fruit to review
+    Suggest(SuggestArgs),
+    /// ??? List audits mechanisms ???
+    Audits(AuditsArgs),
```
Gankra, line 71: Need to go back and figure out what this command actually does, I think it's supposed to be kinda a "manager" command like `git remote` or `rustup toolchains`?

src/main.rs
```diff
@@ -1,3 +1,829 @@
...
+    Debug,
+    Trace,
+}
+
+#[derive(serde::Deserialize)]
+struct Config {
+    // Reserved for future use, if not present version=1 assumed.
+    // (not sure whether this versions the format, or semantics, or...
+    // for now assuming this species global semantics of some kind.
+    version: Option<u64>,
```
Gankra, line 127: I don't really know what this is "for" but it felt right to toss it in.

src/main.rs
```diff
@@ -1,3 +1,829 @@
...
+}
+
+// FIXME: Probably want this to be a tree, and for queries to
+// be keyed off the "current package" but unclear what that would *mean*
+// so for now assume there is a linear chain of overrides. Possible
+// that any "tree-like" situation wants to be an error...
+
+/// All available configuration files, overlaying eachother.
+/// Generally contains: `[Default, Workspace, Package]`
+struct Configs(Vec<Config>);
```
Gankra, line 143: I currently have this setup as a cascade of the 3 sources of config, but in principle package tomls being able to have configs implies a tree-like structure and I have no idea *what* that would mean for cargo-vet's execution, which afaict is currently conceptualized to operate globally and not per-package.

src/main.rs
```diff
@@ -1,3 +1,829 @@
...
+
+/// This is just a big vague ball initially. It's up to the Audits/Unuadited/Trusted wrappers
+/// to validate if it "makes sense" for their particular function.
+#[derive(serde::Serialize, serde::Deserialize)]
+struct DependencyEntry {
+    forbidden: Option<Version>,
+    version: Option<Version>,
+    delta: Option<Delta>,
+    who: Option<String>,
+}
```
Gankra, line 194: This is kinda a hack. Basically you are defining entries as implicitly enums, and the `toml` crate/format doesn't like that. To cope with it, I just have this function represent the union of all the enum variants and then do a post-process validation step (unimplemented, but conceptually doable).

Also since the different files seemed to basically be different subsets of the syntax and I'm already doing post-processing, it seemed reasonable to cram them all together.

The one problem with this definition is the one random place where you introduce `x.*` deps for just one of the files, which might necessitate making some of these into VersionReqs that I also post-process. (These are types from the `semver` crate since it's essentially the canonical source of truth on cargo version syntax, ideally whatever we do should interop with that.)

src/main.rs
```diff
@@ -1,3 +1,829 @@
...
+                    .join(DEFAULT_STORE)
+                    .into_std_path_buf(),
+            ),
+        }),
+        audits: None,
+    };
+
+    let workspace_config = || -> Option<Config> {
+        // FIXME: what is `store.path` relative to here?
+        Config::deserialize(metadata.workspace_metadata.get(WORKSPACE_VET_CONFIG)?)
```
Gankra, line 431: Unclear relative-path semantics

src/main.rs
```diff
@@ -1,3 +1,829 @@
...
+                    e
+                );
+                std::process::exit(-1);
+            })
+            .ok()
+    }();
+
+    let package_config = || -> Option<Config> {
+        // FIXME: what is `store.path` relative to here?
+        Config::deserialize(metadata.root_package()?.metadata.get(PACKAGE_VET_CONFIG)?)
```
Gankra, line 444: Unclear relative-path semantics

src/main.rs
```diff
@@ -1,3 +1,829 @@
...
+    }
+    if let Some(cfg) = package_config {
+        configs.push(cfg);
+    }
+    let config = Configs(configs);
+
+    info!("Final Config: ");
+    info!("  - version: {}", config.version());
+    info!("  - store.path: {:#?}", config.store_path());
+    info!("  - audits: {:#?}", config.audits());
```
Gankra, line 467: All the keys I noticed in the book, any more?

(Also my added "probably useful" optional version key)

src/main.rs
```diff
@@ -1,3 +1,829 @@
...
+
+    let init = is_init(&config);
+    if matches!(cli.command, Some(Commands::Init { .. })) {
+        if init {
+            error!(
+                "'cargo vet' already initialized (store found at {:#?})",
+                config.store_path()
+            );
+            std::process::exit(-1);
+        }
```
Gankra, line 481: Should re-initing a repository do something useful?

src/main.rs
```diff
@@ -1,3 +1,829 @@
...
+    match &cli.command {
+        Some(Commands::Init(sub_args)) => cmd_init(out, &cli, &config, &metadata, sub_args),
+        Some(Commands::Fetch(sub_args)) => cmd_fetch(out, &cli, &config, &metadata, sub_args),
+        Some(Commands::Certify(sub_args)) => cmd_certify(out, &cli, &config, &metadata, sub_args),
+        Some(Commands::Forbid(sub_args)) => cmd_forbid(out, &cli, &config, &metadata, sub_args),
+        Some(Commands::Suggest(sub_args)) => cmd_suggest(out, &cli, &config, &metadata, sub_args),
+        Some(Commands::Diff(sub_args)) => cmd_diff(out, &cli, &config, &metadata, sub_args),
+        Some(Commands::Audits(sub_args)) => cmd_audits(out, &cli, &config, &metadata, sub_args),
+        Some(Commands::HelpMarkdown(sub_args)) => cmd_help_markdown(out, &cli, &config, &metadata, sub_args),
+        None => cmd_vet(out, &cli, &config, &metadata),
```
Gankra, line 499: Any subcommands I missed?

src/main.rs
```diff
@@ -1,3 +1,829 @@
...
+    let store_path = config.store_path();
+
+    let audited_path = store_path.join(AUDITED_TOML);
+    let trusted_path = store_path.join(TRUSTED_TOML);
+    let unaudited_path = store_path.join(UNAUDITED_TOML);
+
+    // Create store_path
+    // - audited.toml (empty)
+    // - trusted.toml (skeleton?)
+    // - unaudited.toml (populated with the full list of third-party crates)
```
Gankra, line 522: unclear what "skeleton" was, but now you've changed these around so need to completely rejig this

src/main.rs
```diff
@@ -1,3 +1,829 @@
...
+    //
+    // If not for deltas, resolving packages would be fairly trivial.
+    // *With* deltas I think we want to have some DAG-like analysis where you start
+    // at the current version and look for a 'delta.to' that matches that version,
+    // and then recursively check `delta.from'
+    //
+    // One thing that's unclear is what should happen if 'delta' hops *over* a version
+    // i.e. if we have `audited.version = 5` and `audited.delta = 3 -> 7', does that
+    // allow us to accept version 7? This is kind of an incoherent state but it seems
+    // plausible with 'trusted' inputs imported from elsewhere!
```
Gankra, line 696: Loose notes on my initial reading of the format and how to "implement" it. Again needs to be rejigged based on book.

Last paragraph covers an "interesting question" about deltas that is probably still relevant.

Gankra (commented):
Notes on initial ambiguities (possibly covered somewhere in the book, but didn't immediately see them)

src/main.rs
Gankra: Ah, this is partially resolved by making the config.toml file! Still a bit unclear what it would mean for a crate to "override" the workspace for the `store.path` though.

README.md
```diff
@@ -0,0 +1,108 @@
+# cargo-vet
+
```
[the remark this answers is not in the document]
Gankra: ah, sample-criteria helps a bit. it's... interesting that it seems to conflate "methodology" and "claims"? Like "reviewed" feels like a very different *kind* of statement to "sound". hmm...

also is there at all a subsetting relationship between "sound" and "secure"? It feels like there should be but I'm genuinely not sure in which direction.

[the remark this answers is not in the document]
Gankra: oh wow actually your changes seem way more wide-sweeping than i expected, ok I really just need to re-read this completely

src/main.rs
bholley: I don't think it's in the writeup at present.

src/main.rs
bholley: Sure

src/main.rs
bholley: Yeah conceptually it's global for the workspace or project. Individual crates shouldn't be able to override it.

src/main.rs
bholley: Probably Cargo.toml?

src/main.rs
bholley: Probably just throwing for now.

src/main.rs
bholley: I recently added `accept-criteria-change`, and per above I don't think `audits` is a thing.

src/main.rs
Gankra: Should we hard error, or just warn (and discard the package's values) if we find both?

Gankra:
(I went on a bit of a markdown generation adventure here because this subroutine is just, actually useful to me elsewhere...)

src/main.rs
bholley: My lean would be to have |struct AuditEntry| and |struct UnauditedEntry| which just map directly to the tables in the toml. We can define an enum over the two of them in situations where we need to abstract across them.

src/main.rs
bholley: This is now documented more comprehensively.

src/main.rs
bholley: Regarding this last question: we should require an additional audited.delta = 5->3 or 5->7. Otherwise, someone could introduce something malicious in version 2, remove it in version 4, and add it back again in version 6.

Gankra: I agree that conceptually it can happen, but in theory "add it back in 6" would have been visible in the claimed delta of `3 -> 7`, no?

...OHH. right right right. I kept conceptually thinking of this as reviewing all the patches/PRs for that delta, but of course the expected/reasonable workflow is to diff the entire repo, and not see the patches themselves. Gross! Nice!

README.md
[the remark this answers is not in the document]
bholley: > default-criteria: "The value may be either a string or an array of strings."
> 
> I assume this implies criteria in audited.toml is also like this?

Yes.

> Are you picturing something like "reviewed, fuzzed, tested"?

I was more imagining a situation wherein Firefox audits for `secure`, and imports audits from another project that does a lot of audits for `sound`. In that case, there could be value in splitting `secure` into two sub-criteria, so that we could skip soundness audits for some crates and just focus on ambient capabilities.

It's quite possible that this is too complex to be useful in practice, so we could also introduce it later as needed, but allowing criteria to be an array struck me as a nice generalization which didn't have any obvious downsides.

> ah, sample-criteria helps a bit. it's... interesting that it seems to conflate "methodology" and "claims"? Like "reviewed" feels like a very different _kind_ of statement to "sound". hmm...

That's a fair point. I had originally named them "correct" and "crypto-correct", but then I got worried that it would feel like an unreasonably high bar (zero bugs in this code), whereas the practical policy would be more about applying first-party-level scrutiny. I'm open to various outcomes here, though since it doesn't affect the implementation we can probably punt on it for now.

> also is there at all a subsetting relationship between "sound" and "secure"? It feels like there should be but I'm genuinely not sure in which direction.

As written, "sound" is almost a strict subset of "secure", except for the matter of degree (i.e. "secure" makes some allowances for discretion, whereas "sound" doesn't). I think we want the discretion in "secure" for Firefox's use-case, but I thought I'd get a lot of grief from soundness purists by introducing similar discretion in "sound".

Gankra:
updates:

# rejigged stuff to the shape of the new format 

I have soft-vetoed your proposal for forbidden/unaudited versions to have a bespoke syntax. I have turned them into [proper cargo VersionReqs](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html), to continue the story of "using the existing concepts". This is actually mostly a superset of what you proposed *except* it means `x.y.z` is actually `^x.y.z`, whereas I think you intended it to be `=x.y.z`. If you think that's a bit wonky, I agree. I think there's a few options here. The simplest is to just forbid bare `x.y.z` as a post-process step, so it's always clear which fields are VersionReqs, and which are Versions.

One benefit of going Full VersionReq is that there is now no need to have multiple entries for unaudited libs, because VersionReqs actually let you infinitely recursively concat them with commas. This... may be an accidentally-quadratic scaling issue in the longterm. I need to think about it more (but that would be just as true if you had separate entries too, right?).

(Also I almost certainly just missed some random stuff, a lot of changes...)

# sketched out a resolution algorithm for deltas. 

This is not properly tested and will probably fall over if you look at it funny, but I think this is a vaguely sound / bounded algorithm, assuming no degenerate deltas like "x -> x" or "x+1 -> x" which could introduce infinite loops. main downside of the current approach is that i'm not "remembering" that a version has already been visited, so if you get to a version again via a longer path, you'll re-explore its whole subtree again. This can be fixed by just maintaining a HashSet of all visited versions, but I didn't want to overthink it right now.

# some gardening and thinking about editing

I've introduced LinkedHashMap in a lot of places as a vague concession to trying to "preserve" manual orderings of the file's contents when doing something like `vet certify`. The naive design here is to take in a `toml`, parse it into a struct, make a small edit to the struct, and then reserialize the whole struct to a fresh file. The official toml crate has no notion of comments or formatting, so this would blast all comments into the aether and basically `fmt` your file. I could see this being tolerable if hand-edits are considered rare, and message strings are explicitly part of the structure (they are as I see!).

Doing better than this involves using the [toml_edit](https://killercup.github.io/cargo-edit/toml_edit/) crate, which is what [cargo-edit](https://github.com/killercup/cargo-edit/) uses. This would be More Annoying and Fiddly, as we would have to more carefully navigate the toml and poke at the thing we want to change.

I am avoid going down the toml_edit path for now, since it's extra complexity that doesn't matter, and I'm reasonably confident I can retrofit it in later if we *really* care.

Gankra:
Next on my docket is to slap Reqwest in here and implement the networking logic.

bholley:
> I have soft-vetoed your proposal for forbidden/unaudited versions to have a bespoke syntax.

I appreciate this pushback. :-)

> I have turned them into [proper cargo VersionReqs](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html), to continue the story of "using the existing concepts".

The use-cases for VersionReqs are different enough that I'm not sure it makes sense to adopt them wholesale. I don't think we should support either `~` or `^` (saying "I audited this crate and any future minor versions" defeats the point), at which point requiring a preceding `=` for all precise versions just serves as clutter. It would also create consistency issues with delta syntax unless we introduced `=` there as well, and `=1.2.1 -> =1.2.2` doesn't feel right.

That said, I totally agree that we should jettison my interval proposal and lean on the VersionReqs syntax for interval and star-matched versions.

> One benefit of going Full VersionReq is that there is now no need to have multiple entries for unaudited libs, because VersionReqs actually let you infinitely recursively concat them with commas.

I'm actually wary of adopting the comma syntax, because I think it's easier from VCS perspective to have each audit (or unaudit) have its own block.

> This... may be an accidentally-quadratic scaling issue in the longterm. I need to think about it more (but that would be just as true if you had separate entries too, right?).

I'm not totally sure what you mean. You're just talking about the algorithm getting expensive to run as either the `audits` or `unaudited` tables get very large? I suspect that won't be an issue in practice.

> I am avoid going down the toml_edit path for now, since it's extra complexity that doesn't matter, and I'm reasonably confident I can retrofit it in later if we _really_ care.

Yes I think this is the right approach. The original write-up leaned on comments more but for now I think we should see if we can get away with auto-formatting. We should probably also introduce a `cargo vet format` or something to trigger the reformatting without certifying any audits (so that people can format after manually editing the file, which is sometimes necessary for the more bespoke use-cases that `cargo vet certify` won't support. We should probably also auto-format `config.toml`.

Gankra:
So there's only two places where VersionReqs should ever show up:

* unaudited
* forbidden

They shouldn't ever show up in "this code is good" audits entries or deltas, because that doesn't make sense. Like ok it *kind of* makes sense but it's really weird to say "I reviewed 1.0.*" or "I reviewed 1.0.0 -> 1.0.*" and that is not actually a claim you can make because versions do Time Travel:  `cargo publish` lets you publish to "the past" or basically any version number, so you can never confidently say "I have reviewed every version in this range". Even reasoning about "well there's no versions between 0.1.2 and 0.1.3 is *wrong* because Cargo [supports "0.1.2-prerelease"](https://docs.rs/cargo_metadata/latest/cargo_metadata/struct.Version.html) and the like.

If you're trying to express I own this code actually, I trust all versions, then I guess that's *maybe* fine, but it's probably better to still just say that for ever individual release you care about. Any range or \* syntax is pretty aggressive (will you own it forever? did you always own it? do you really want your trust to implicitly extend infinitely into the past and future? even knowing the time travel problem?)


------

On unaudited:

For unaudited, I don't particularly see a reason to want to ever distinguish between different unaudited entries for the same crate. i.e. I would expect them to all just be "haven't gotten to it yet, *shrug*. The one distinction I could maybe see is wanting to separate out a version as like "woah this change was a lot, this needs more scrutiny, we're punting on this for now"... but idk if that's a real distinction worth making? No matter what, these entries are just "shrug" in my mind.

For this reason I expect it might be common to end up with an unaudited entry that's basically 

```toml
[mylib] 
version = "=0.0.1, =0.0.2, =0.0.3, ..."
```

which is to say every time you update it and continue to punt on reviewing it, you are just concatenating the new version to the end of the list, because having 50 entries that are all "yep, still haven't done it" is not terribly useful.

Now regardless of whether you have one comma-separated mega-value or actual separate entries, if enough of these "exact versions" pile up then it can become a bit of a perf hazard for the resolver, because every time we "walk" to another version through the deltas, we need to check if that version is in the unaudited list, which if done via searching a list linearly, is Not Great. The "fix" for this is to dump all the entries into a HashMap (which would require me "destructuring" the version-req into its parts, blech)... but this only works for exact versions. If someone naively puts a ton of `>` or `*` entries in there then I either need to build like a range-find structure or fall back to linear!

Anyway that's mostly "deal with it when it happens" and not "oh my god existential".

As for Time Travel with unaudited: it's a *little* scary that you can time-travel a new version into one's trust base by abusing their unaudited ranges, but this *mostly* isn't a problem because:

A) when using ranges they're already basically saying "I really don't care right now"
B) versions mostly only go up. I suppose you *could* but badness in a transitive dependency and then have a direct dependency require a lower version? I'm not actually 100% sure how Cargo would react to this!
C) unlike audits, these ranges don't get "exported" to third-parties who trust you. so time-travel attacks are only really "direct".

------

On forbidden:

In this case I can expect you to actually want multiple entries, because version A might be forbidden for a completely different reason than version B. That's just a project with a lot of quality control problems :)

As for Time Travel with forbidden: it's basically not a concern at all. At worst you'll be overbroad and ban some good releases, but that's "fine". That said overbroad forbids are another reason there should probably be a way for a local audits file to override forbids of a foreign file to say either "they're wrong, the range is overbroad" or "they're being pedantic/excessive, this is fine".

(I'm not sure if the version policy stuff you have sketched out handles this, I haven't gotten to that yet, if it's explained in the book at all)

bholley:
> So there's only two places where VersionReqs should ever show up:

Ah ok, I mistakenly thought we'd be applying VersionReqs to everything.

> They shouldn't ever show up in "this code is good" audits entries or deltas, because that doesn't make sense.

Agree with this.

> On unaudited:
> 
> For unaudited, I don't particularly see a reason to want to ever distinguish between different unaudited entries for the same crate. i.e. I would expect them to all just be "haven't gotten to it yet, _shrug_. The one distinction I could maybe see is wanting to separate out a version as like "woah this change was a lot, this needs more scrutiny, we're punting on this for now"... but idk if that's a real distinction worth making? No matter what, these entries are just "shrug" in my mind.

Generally agree, which is why `unaudited` doesn't support deltas.

> 
> For this reason I expect it might be common to end up with an unaudited entry that's basically
> 
> ```toml
> [mylib] 
> version = "=0.0.1, =0.0.2, =0.0.3, ..."
> ```
> 
> which is to say every time you update it and continue to punt on reviewing it, you are just concatenating the new version to the end of the list, because having 50 entries that are all "yep, still haven't done it" is not terribly useful.

This won't be the case, because you'd generally prune the old ones after each update (`cargo vet` warns about unnecessary `unaudited` entries).

> As for Time Travel with unaudited: it's a _little_ scary that you can time-travel a new version into one's trust base by abusing their unaudited ranges, but this _mostly_ isn't a problem because:
> 
> A) when using ranges they're already basically saying "I really don't care right now" B) versions mostly only go up. I suppose you _could_ but badness in a transitive dependency and then have a direct dependency require a lower version? I'm not actually 100% sure how Cargo would react to this! C) unlike audits, these ranges don't get "exported" to third-parties who trust you. so time-travel attacks are only really "direct".

Agree that ranges in `unaudited` are risky. I was on the fence about adding them, and eventually decided in favor to support the use-case described [here](https://github.com/bholley/cargo-vet/blob/7ae734f7dfdc4e1334f59f48a8a9ccf229718e4d/book/src/platform-specific.md). But maybe it's better to just not support them.

> On forbidden:
> 
> In this case I can expect you to actually want multiple entries, because version A might be forbidden for a completely different reason than version B. That's just a project with a lot of quality control problems :)

I think if you're forbidding different versions for different reasons, you want different entries.
 
> As for Time Travel with forbidden: it's basically not a concern at all. At worst you'll be overbroad and ban some good releases, but that's "fine". That said overbroad forbids are another reason there should probably be a way for a local audits file to override forbids of a foreign file to say either "they're wrong, the range is overbroad" or "they're being pedantic/excessive, this is fine".

Yes, agreed.

So I think my takeaway from this is that `forbidden` is the only thing that really needs fancy range stuff, and also likely to be pretty rare. So maybe we should just drop the extra from `unaudited` for now (which would allow `version` to have the same data time for both Audit Entries and Unaudited Entries), and use a full VersionReqs for `forbidden`.

Gankra:
I think if we can make the workflow as painless as [cargo insta](https://github.com/mitsuhiko/insta) for the "unaudited dep updated, you want to just bump the number right?" case, then it's not a problem to churn unaudited deps a lot. 

Basically with insta it enters an interactive state and asks you to accept/reject each change individually. Generally you expect every change you see so you spend about 10 seconds mashing "A".

bholley:
> I think if we can make the workflow as painless as [cargo insta](https://github.com/mitsuhiko/insta) for the "unaudited dep updated, you want to just bump the number right?" case, then it's not a problem to churn unaudited deps a lot.

I actually don't want to make it too easy to just bump the deps in the unaudited list. The preferred thing to do in that situation is to add a delta audit, which protects you from attacks going forward. The original concern about it being a hassle to update dependencies was mostly about unused platforms, but now we have a better way to handle those.

PR #2, opened 2022-04-12
cargo-vet skeleton mvp

Gankra:
hacky hacky hacky


Gankra:
Went through the book again and got schemas/apis to match things to the best of my understanding.

Lots more explicit TODOs added for stuff I don't do yet.

Error handling still a mess (experimenting with some new stuff, feeling out what approaches I like, so kinda doing random stuff all over the place. Also mildly enticed to mess around with [color-eyre](https://github.com/yaahc/color-eyre) but not sure it's right for this project...).

Gankra:
Also see https://github.com/Gankra/cargo-vet#cargo-vet for rendered CLI docs

tests/snapshots/test_cli__short-help.snap
```diff
@@ -0,0 +1,39 @@
...
+                                       [possible values: off, error, warn, info, debug, trace]
+        --workspace                    Process all packages in the workspace
+
+SUBCOMMANDS:
+    accept-criteria-change    Accept changes that a foreign audits.toml made to their criteria
+    certify                   Mark `$package $version` as reviewed with `$message`
+    diff                      Yield a diff against the last reviewed version
+    fetch                     Fetch the source of `$package $version`
+    fmt                       Reformat all of vet's files (in case you hand-edited them)
+    forbid                    Mark `$package $version` as unacceptable with `$message`
```
bholley, line 35: This isn't a thing. The terminology has changed to `violation`, and there's no CLI command for it: https://bholley.net/cargo-vet/design-choice-faq.html#why-is-there-no-helper-for-recording-violation-entries

Gankra: Ah yeah, I saw everything got changed to "violation" and fixed that, but wasn't sure about the command (hard to find the absence of something when skimming docs you're rereading for the 3rd time).

Minor question: how do you refer to a failure-as-a-result-of-violation? Before I could say "found a forbidden package" or something, "found a violated package" is... not right.

bholley: "packaged marked as violating CRITERIA_FOO". "problematic package" would probably do in a pinch.

Gankra:
Updates:

* removed Vetted.lock
* removed forbid command
* kinda-WIP pivot to using registry cache (problems now shifted from "we explode on multiple versions of the same package" to "we completely fail to actually pull down anything you ask about unless it happens to be in your cargo.lock/cargo.toml"

Also I'm still not implementing the "find the closest audited/allowed package to the current version" logic I mentioned in #13, so anything other than a "0.0.0 => now" will uh... Not Work.

That's next on my plate.

Gankra:
NB `suggest` is still *pretty* heavyweight since just... diffing 200 crates is still just a ton of disk access and non-trivial work, even for `git diff`!

I know this "smoothes out" over time as you mark things as suggest=false and just review them and maybe we cache results and whatnot, but it's... not a great first run UX? (and also useless to do in a CI vet?)

Gankra:
Or wait, hold on there's kind of 2 suggests isn't there?

* suggest what to fix when `vet` fails
* suggest what to audit to just clear up your unaudited backlog

no?

Gankra:
Pushed up: new `vet` resolution algorithm that is criteria-aware and maps in foreign criteria. current caveats:

* local dependency_criteria not yet implemented
  * need to make sure the array is properly sorted for dependency resolution first
* foreign dependency_criteria not yet implemented (see #16)
* foreign criteria are mapped immediately, and all resolution is performed in local terms
  * this can result in "problem b" in https://github.com/bholley/cargo-vet/issues/16#issuecomment-1099133351
* policy is not actually checked 
  * I don't think it's tricky/interesting, just haven't looked at it
* implies is not actually applied yet
  * literally forgot it exists until I started writing this comment lol

Most of this is noted in TODOS

src/main.rs
```diff
@@ -1,3 +1,1827 @@
...
+    }
+    fn is_empty(&self) -> bool {
+        self.0 == 0
+    }
+}
+
+fn cmd_vet(out: &mut dyn Write, cfg: &Config) -> Result<(), VetError> {
+    // Not sure which we want, so make it configurable to test.
+    // Determines whether a delta must be == unaudited or just <=
+    let unaudited_matching_is_strict = true;
```
bholley, line 1056: Intention is for it to be strict.

Gankra: I can't remember where I wrote this down but I'm not convinced this is the right choice. If you say "unaudited=v3", you're basically saying "i have punted on this up to v3". So if someone comes along with an audit from v2->v5, why *would* you reject that as a proof of v5? v3 isn't at all special to you, right? 

We just want it to be a strict version number to force you to audit whenever you update the dep with no proof at all, right? The main thing you need to ensure is strict equality for "delta-less" acceptances (i.e. for when cur_version == unaudited).

Gankra:
Pushed up dependency_criteria / dependency-order resolution. Checked in some test data just to make sure things weren't completely broken... this is getting complex enough that I desperately need to write the testing framework.

Gankra:
NB while I'm still thinking through foreign dep_criteria, I am currently unconditionally setting that stuff to "default" (and aborting if custom criteria is set). More work to do here, but should vaguely do the right thing in defaulted bijective cases.

src/main.rs
bholley: Consider the scenario where v1 had a vulnerability, which was fixed in v2, and you're using v3 (via unaudited). An attacker could theoretically reintroduce the vulnerable code in v4. In this situation, the v1->v4 diff would be clean, but the v3->v4 diff would introduce a vulnerability.

I don't claim that this scenario is particularly likely, but I think `unaudited_matching_is_strict=false` is a sort of surprising optimization that's not 100% sound.

bholley: Intuitively, I think having something in unaudited means "I am going to take the risk of assuming this version is safe", not "I will assume this and all prior versions are safe".

Gankra: alright, fair enough. I think it's still fine to leave in here to mess around with

PR #18, opened 2022-04-14
fix markdown-help and add simple integration test

Gankra:
see tests/test-project/audits.toml for the current tests (no fail tests, but I locally messed them up to check that it broke right too...)


Gankra:
In the longterm I'll factor out the resolution stuff so that I can feed dummy data in without a full project, but this lets us just try random stuff and see what happens

Issue #19, opened 2022-04-15
questions on policies


Gankra:
> By default, cargo vet checks all transitive dependencies of all top-level crates against the default criteria for all platforms. In some situations, you may be able to reduce your workload by encoding your requirements more precisely, such as:
>
>  Using different criteria for different top-level crates. Your workspace might contain both a production product and an internal tool, which might require different levels of audits for their dependencies.

I am not current aware of a mechanism for "using different criteria for different top-level crates". The design I have seen so far actually makes it *impossible* to talk about top-level crates, because they are definitionally not-third-party, and therefore completely ignored by all operations!

So policies let you set the default criteria everything must pass, but how *exactly* do you specify that a *crate* has a different policy?

# You can't specify that a crate should pass higher criteria

You can use dependency_criteria on *audits* to enforce that a dependency *edge* has a different policy, and by default an audit just projects its claimed criteria to all edges, so a load-bearing audit is *kind of* requiring that the crate meets that policy... but in a weird wobbly way that can be routed around. Like even if you have super-awesome-secure delta-audit, as long as your deps meet that criteria we'll happily walk through that audit even if this delta depends on weaker criteria because we're currently silently intersecting things and the final arbiter of if the result is "good enough" is still the `policy`.

# You can't specify that a crate can have lower criteria

The problem I see is that (as described in the book right now) your global policy is an effective *floor* on the criteria. If you say policy=super-awesome-secure, then it doesn't matter if you're fine with barely_used_dependency having criteria=not-obviously-a-virus, the policy will reject the crate, even if all the edges to it say it's ok for it to only be not-obviously-a-virus.

The only way to allow anything to have not-obviously-a-virus is to make the policy not-obviously-a-virus, but then you're making the checking on the root crates even weaker!

----------------

I'm not sure if you just haven't sketched this part out, or if it somehow made more sense to you when you were visualizing things top down, but I don't really know how to use this system right now.

bholley:
> I am not current aware of a mechanism for "using different criteria for different top-level crates". The design I have seen so far actually makes it _impossible_ to talk about top-level crates, because they are definitionally not-third-party, and therefore completely ignored by all operations!
> 
> So policies let you set the default criteria everything must pass, but how _exactly_ do you specify that a _crate_ has a different policy?

[...]

PR #20, opened 2022-04-15
policy enforcement and better diagnostics

Gankra:
See #19 for the issues.

I have temporarily checked in the test failure this introduces so you can see what breaks. Policy as currently specified makes it impossible to have "meh" dependency edges without weakening the global policy.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -1241,6 +1241,18 @@ fn cmd_vet(out: &mut dyn Write, cfg: &Config) -> Result<(), VetError> {
     let criteria_mapper = CriteriaMapper::new(&audits.criteria);
     let all_criteria = criteria_mapper.all_criteria();
     let no_criteria = criteria_mapper.no_criteria();
+    let policy = config
+        .policy
+        .criteria
+        .as_ref()
+        .map(|c| criteria_mapper.criteria_from_list(c.iter().map(|s| &**s)))
+        .unwrap_or_else(|| criteria_mapper.default_criteria().clone());
+    let _build_and_dev_policy = config
+        .policy
+        .build_and_dev_criteria
+        .as_ref()
+        .map(|c| criteria_mapper.criteria_from_list(c.iter().map(|s| &**s)))
+        .unwrap_or_else(|| policy.clone());
 
     // This uses the same indexing pattern as graph.resolve_index_by_pkgid
     let mut vet_resolve_results = vec![no_criteria.clone(); graph.resolve_list.len()];
@@ -1476,8 +1488,8 @@ fn cmd_vet(out: &mut dyn Write, cfg: &Config) -> Result<(), VetError> {
                 }
             }
         }
-        // TODO: now verify validated_criteria matches our policy
-        let passed_policy = !validated_criteria.is_empty();
+        // TODO: distinguish between build_and_dev somehow?
+        let passed_policy = validated_criteria.contains(&policy);
 
         if passed_policy {
             // hooray, we win! record the result in vet_resolve_results
--- a/src/main.rs
+++ b/src/main.rs
@@ -1185,7 +1185,7 @@ impl CriteriaSet {
     fn contains(&self, other: &CriteriaSet) -> bool {
         (self.0 & other.0) == other.0
     }
-    fn _is_empty(&self) -> bool {
+    fn is_empty(&self) -> bool {
         self.0 == 0
     }
 }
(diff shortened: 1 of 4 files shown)


Gankra:
New push massively fleshes out the information we gather during a vet, and implements policy.

Here is what a failure looks like now:

```
 cargo vet
Vetting Failed!

2 unvetted dependencies:
  test-project:0.1.0
    clap:3.1.8
      bitflags:1.3.2

recommended audits:
  [TODO]

```

Gankra:
Success now looks like this:

```
Vetting Succeeded (6 fully audited 1 partially audited, 98 unaudited)
```

I personally think it's interesting to distinguish between "deltas terminated in unaudited" as "partially audited", but ymmv.

Gankra:
Hmm still need to bang on this unaudited checking a bit more...

Gankra:
Ok I went kinda off the rails for a bit with that last commit because I think the way I'm testing resulted in temporary Extreme Cursedness where vet was somehow using the parent project's Cargo.lock(!?!?!?) and so all everything was exploding because I was now more correctly validating unaudited deps.

ANYWAY, we now spit out this:

```
Vetting Succeeded (7 fully audited 2 partially audited, 96 unaudited)
  warning: some dependencies are listed in unaudited, but didn't need it:
    libc:0.2.123
    log:0.4.16
```

Gankra:
So the current approach is now roughly this:

* validate all the third-party packages bottom-up:
  * if there is a directly-unaudited entry for the package's current version, don't treat that as a validation, but instead note it
  * when checking deps, version comparison now explicitly falls back to whether the dep is directly-unaudited (allowing us to notice when that was needed)
  * if a dep check fails, log it as "to blame" if we end up failing completely
  * if we didn't actually find any audit path for a package, and it's directly-unaudited, require our deps to have the default criteria to make sure nothing gets missed (I think this is the only way we'd notice a brand new unknown dep of an unaudited crate)
    * if this guard fails, mark the package as "failed", undoing its directly-unaudited status for dep checks (also log "to blame")
    
* validate all first-party packages bottom-up:
  * check that their deps satisfy the policies
  * if not, mark the first-party package as failed, log the failing deps as "to blame"

* gather stats 

* if any workspace packages failed (which should occur if *anything* "failed")
  * recursively print the "to blame" logs, printing out a dep tree

Gankra:
This design works well for the "everything is the default" workflow but will kinda freak out and point a lot of random fingers in more complex situations, because it's very difficult to identify whether a failed dep-check on an audit was "load-bearing" or just some dead path that wasn't going to pan out anyway.

I started vaguely gesturing at ways to approach this in the implementation, where we now keep exploring even if the intersection becomes empty. I think we could maybe do a better job if we continue to traverse down paths with failed dep checks "speculatively" and note if they *would* have worked but... it's complicated.

src/main.rs
```diff
@@ -249,13 +249,48 @@ struct CriteriaEntry {
...
+/// If this sounds overwhelming, don't worry, everything defaults to "nothing special"
+/// and an empty PolicyTable basically just means "everything should satisfy the
+/// default criteria in audits.toml".
 #[derive(serde::Serialize, serde::Deserialize)]
 struct PolicyTable {
+    /// Default criteria that must be satisfied by all *direct* third-party (foreign)
+    /// dependencies of first-party crates. If satisfied, the first-party crate is
+    /// set to satisfying all criteria.
+    ///
+    /// If not present, this defaults to the default criteria in the audits table.
```
bholley, line 272: In the situation where the first-party crate is not a root, I think we want it to default to the parent's criteria. Only root nodes should default to the default criteria.

Example here is: I want all of geckodriver to use `safe_to_run_locally`, and just want to specify that once (at the top-level geckodriver crate) rather than duplicating it around the first-party nodes of the subtree.

src/main.rs
```diff
@@ -249,13 +249,48 @@ struct CriteriaEntry {
...
     criteria: Option<Vec<String>>,
+
+    /// Custom criteria for a specific first-party crate's dependencies. It's nonsensical
+    /// to define criteria between two first-party crates, so you can only name third-party
+    /// dependencies here.
+    ///
+    /// Any dependency edge that isn't
```
bholley, line 279: Missing the rest of the text here

src/main.rs
Gankra: Ok so then it *does* make sense to have special rules between first-party crates, as just a way to specify that a particular sub-first-party should have such and such policy..? I'm not sure how that works if a first-party has multiple reverse-dependencies... maybe first-parties *do* need to be processed top-down?

bholley: So basically, I'm envisioning two steps: a bottom-up traversal to compute |self| and |subtree|, and then a top-down traversal to check the policy and compute the information needed for suggest.

Gankra: policies don't "descend" into third-party transitive deps right? I don't really understand what that would mean. I guess you could take the union of your reverse-dependencies..?

Gankra: basically my confusion is that the computation of |subtree| for a first-party crate *depends* on policy, so I don't know how you handle first-party crates bottom-up? And then if you let policy descend into third-parties then I am even more confused.

Issue #30, opened 2022-04-21
safe_to_build has unclear purpose


Gankra:
Initially it seemed like an obviously useful distinction, but the more I think about it the less it makes sense. Well, it doesn't make sense in two ways:

* the way it's applied in `policy` seems wrong
* even granting that, it's unclear what relevance the distinction has

So tackling the first case, we have this:

> ### the `policy` Table
>
> ... 
>
> #### Build dependencies
>
> Same as the above, but applied to build-dependencies.
>
> For top-level crates, defaults to `safe_to_build`.

Requiring build-dependencies to be merely "safe to build" doesn't really make sense, because they are *run* as part of your build. Either you want "safe_**for**_builds" or just "safe_to_run_locally". I don't think rust/cargo really has much of a sandboxing story for build-deps, so it's unclear to to me how "safe for builds" would be materially different from "safe to run locally"?

This brings us to the larger issue: under what circumstances would it be noteworthy that a project is *merely* safe to build? This would seem to imply a strange situation where you are building in a privileged environment and then deploying to a **less** privileged one where it's fine for the program to do whatever? This is the reverse of my expectation -- that for reproducibility you might build some software in a weird docker image or something that provides some incidental sandboxing, but then you pull it out to distribute or deploy to production/users where safety requirements are greater.

I think the only situations I can imagine where you care about this are:

* you are specifically doing some static analysis which requires compiling the project to get some artifacts to analyze and nothing more (niche but ok... although not even crater "only" builds it, it runs the test suite too, I believe?)

* you are specifically only making the distinction so that someone can specifically audit a build.rs on its own, and then every future version that doesn't touch the build.rs can get a trivial delta to indicate "this part's still fine"... but then, you really want at least "safe to run locally" so you're going to have a delta chain of that too, and in that chain you can also see that the build.rs isn't touched so you're not even worried about it still being ok? right?

Gankra:
Ah actually, ok one application of "safe to build" would be specifically running rust-analyzer, which is why vscode will prompt you for "hey do you trust this" so that it doesn't actually get your machine exploited for just looking at the source code... but unless vscode/rust-analyzer is materially slurping up these definitions (and somehow understanding this local checkout is "the real mccoy", it's unclear to me how these audits affect that calculus?

bholley:
You're totally right. It feels like such an elegant model until you look at it carefully. :-)

So I had two basic motivations with this. The first was the build-dependencies thing, which as you note, is fundamentally misconceived.

[...]

PR #31, opened 2022-04-21
More tests + Cleanups + Comments

Gankra:
Nothing in here taxes "policy" yet because I'm still rolling it over in my head. I'm also deferring introducing builtins or breaking up build/dev stuff since I have some concerns in #30 that may change the design (and it doesn't really affect the core resolver/suggest functionality anyway, which I am regarding as top priority).

### DIFF
--- a/src/resolver.rs
+++ b/src/resolver.rs
@@ -659,6 +659,33 @@ pub fn resolve<'a>(
         if !found_any_path {
             // If we didn't actually find any paths and we're directly unaudited,
             // make sure our deps make sense (has default criteria)
+            //
+            // A bit of a curious semantic corner case: if there was no way to even
+            // plausibly audit you, does it make sense to care about if your dependencies
+            // are messed up too?
+            //
+            // Consider the case where both "X" and "dep_of_X" are completely new and
+            // unaudited (or just both updated so we have a broken chain on both). They are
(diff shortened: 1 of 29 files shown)


src/resolver.rs
```diff
@@ -659,6 +659,33 @@ pub fn resolve<'a>(
...
+            // problems (so dep_of_X might be fine, but we freak out about it because X
+            // is broken and don't know what to check it against).
+            //
+            // The following implementation errs on the side of under-reporting -- if
+            // X has no path *and* isn't directly_unaudited, then it's children will
+            // implicitly be assumed to be fine until we get more information. Fixing
+            // X may suddenly introduce "more work to do", making it potentially difficult
+            // estimate just how broken your current audit tree is.
+            //
+            // See "mock-simple-no-unaudited" for how we currently react to this situation.
```
Gankra, line 688: (I spent weirdly too long thinking about this problem... I noticed it in my testing and was freaking out that I had a nasty problem in my algorithm but once I figured it out and went to "fix" it, I realized that it might actually just be the right choice, even if it's a bit weird... although maybe I should just add the "failed" special-case I hint at?)

bholley: This is the same issue we were discussing a few days ago (the (1) in my suggest sketch). I do think we should probably do the approximation, but that can be a followup — filed #32.

PR #35, opened 2022-04-22
Completely rework the resolver algorithm into repeated/reversible path-connectedness checking

Gankra:


src/tests.rs
```diff
@@ -1087,3 +1081,89 @@ fn mock_simple_delta_to_too_weak_full_audit() {
     let stdout = String::from_utf8(stdout).unwrap();
     insta::assert_snapshot!("mock-simple-delta-to-too-weak-full-audit", stdout);
 }
+
+// TESTING BACKLOG:
+//
```
Gankra, line 1086: This part was a real relief to write out, but also oh my god there's so many more tests to write and so many weird corner cases aaaaaa

Gankra:
As discussed, a few things are stubbed out right now, because I need to go back and rework the core criteria/front-end semantics to catch up with you, and those changes affect details of suggest and whatnot.

Gankra:
My next pass will be catching up to the book.

After that I can finally "finish" suggest (although a lot of the heavy lifting is actually done in this PR, it's mostly stitching together the results and doing the fetches/diffs.

PR #36, opened 2022-04-23
Catchup On The Book

Gankra:
This factors out the file format to its own module, and then pulls a lot more stuff in line with what the book says. Also introduces a ton more serde magic for defaults.

A lot of testing stuff had to be rejigged because the book has kind of changed a lot of defaults from "super late bound" to "super early bound", and the test harness was abusing the late-boundness to elide things that were the default.

Also now the test suite is a little weird in that it's kinda explicitly blanketing over the existence of the builtins. However the "test project" has been updated to use the builtins where it makes sense, so that's good.

Any kind of build_and_dev distinction is still largely ignored and stubbed out.

Also build-targets and dev-targets are still build-and-dev-targets on the assumption that those still being split up was a typo (feel free to convince me otherwise).

### DIFF
--- a/src/format.rs
+++ b/src/format.rs
@@ -94,6 +94,15 @@ pub struct AuditsFile {
 impl AuditsFile {
     pub fn validate(&self) -> Result<(), VetError> {
         // TODO
+        // * check that each CriteriaEntry has 'description' or 'description_url'
+        // * check that all criteria are valid in:
+        //   * CriteriaEntry::implies
+        //   * AuditEntry::criteria
+        //   * DependencyCriteria
+        // * check that all 'audits' entries are well-formed
+        // * check that all package names are valid (with crates.io...?)
+        // * check that all reviews have a 'who' (currently an Option to stub it out)
+        // * catch no-op deltas?
         Ok(())
     }
 }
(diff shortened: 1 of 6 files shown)


Gankra:
(this is a NFC in sense that it basically leaves us in the same place feature-wise, but now I think I'm caught up on "things I already implemented but were changed")

src/format.rs
```diff
@@ -0,0 +1,392 @@
...
+    pub criteria: String,
+    #[serde(flatten)]
+    pub kind: AuditKind,
+}
+
+#[derive(serde::Serialize, serde::Deserialize, Debug)]
+#[serde(untagged)]
+pub enum AuditKind {
+    Full {
+        version: Version,
```
Gankra, line 139: also the fact that this thing actually *works* is completely blowing my mind, i fucking love serde

src/format.rs
```diff
@@ -0,0 +1,392 @@
...
+/// since first-party crates are defined by "not on crates.io").
+///
+/// Because first-party crates are implicitly trusted, really the only purpose of this
+/// table is to define the boundary between a first-party crates and third-party ones.
+/// More specifically, the criteria of the dependency edges between a first-party crate
+/// and its direct third-party dependencies.
+///
+/// If this sounds overwhelming, don't worry, everything defaults to "nothing special"
+/// and an empty PolicyTable basically just means "everything should satisfy the
+/// default criteria in audits.toml".
```
bholley, line 273: This isn't quite right, see https://bholley.net/cargo-vet/specifying-policies.html

src/format.rs
```diff
@@ -0,0 +1,392 @@
...
+#[derive(serde::Serialize, serde::Deserialize)]
+pub struct PolicyTable {
```
bholley, line 275: Shouldn't this be PolicyEntry?

Gankra: Yes so I have this entire time been operating on the understanding that there is One PolicyTable Defining Defaults and not that it's `[policy.my_crate]` like with audits. I don't know if you changed this at one point, or if I have just been missing that fact the whole time, but the docs don't directly state that policy tables are keyed on a crate? It's just implied by the prose talking about "the" crate.

This a pre-existing error(?) though, this PR doesn't change that. Policy stuff will be cleaned up in the next round since policy and suggest are very fundamentally intertwined.

src/format.rs
Gankra: Oh yes, this comment is out of date, but I need to rework policy completely anyway.

bholley:
> Also build-targets and dev-targets are still build-and-dev-targets on the assumption that those still being split up was a typo (feel free to convince me otherwise).

It was a typo, I just pushed a fix.

PR #38, opened 2022-04-27
WIP reworked error reporting and suggest

Gankra:


Gankra:
Ok this appears to be in reasonable shape and largely handle everything. I ended up implementing a bunch of the "unification" logic I wanted to put off because I wasn't going to get decent results in any situation without it.

Gankra:
tomorrow I'll basically just be writing dozens of tests to check that this actually at all does what it should.

Gankra:
Oh one minor thing this is missing from the suggest output is actually sorting the results... bigger fish to fry.

Issue #29, opened 2022-04-20
`cargo vet inspect` and `cargo vet diff` should work anywhere


bholley:
These don't conceptually depend on the specific instance, and so it would be nice if people could invoke them without having to instantiate a dummy project.

Gankra:
i fixed this a while back

bholley:
I still get:

```
bholley@atomic ~ $ cargo vet diff alsa 0.4.2 0.4.3
[ERROR] 'cargo metadata' failed: `cargo metadata` exited with an error: error: could not find `Cargo.toml` in `/Users/bholley` or any parent directory
```

PR #60, opened 2022-05-02
completely rework fetching and caching infra

### LINKED ISSUE #55: properly implement fetch
Pull down the package via the REST API: crates.io/api/v1/crates/crate_name/vers/download (e.g. https://crates.io/api/v1/crates/diesel/1.3.3/download)

untar/zip it to tmp, maintain our own package cache or something

TODO: 

* figure out what the compression format is
* figure out if it's tarred our what

Gankra:
Fixes #55 

* We now have a single Cache struct which can theoretically acquire a global lock (not actually implemented, lockfiles can be very cursed and make things worse if you're not careful...)
* Fetches now check the cargo registry, then check our fetches, and then finally download the file
  * Check looks for a .cargo-ok file which is written only after a complete successful fetch+decompress
* Downloading the crate vaguely mimicks Cargo but with less sanity checks
  * GET the url
  * assume it's a tar.gz file and unpack to our cache
  * write the .cargo-ok file
* The diff_cache is now only loaded and written once per execution, instead of repeatedly

Since we no longer run 'cargo' during fetch/diffing, everything seems significantly more responsive. Also now *everything* feeds into this common machinery so there's less random garbage in each command since most of the details are handled by the Cache.

This doesn't handle #29.


Gankra:
NB: it's very easy for this kind of thing to accidentally rely on existing cruft I had on my system, or have platform-specific issues. I've tried my best to clean things out and check that things work, but, ymmv.

PR #84, opened 2022-05-19
Misc refactors/fixes/features

Gankra:


Gankra:
worth noting that the implementation of certify is still a bit weird because you don't specify how to pick what criteria you're specifying. right now i just use config.default_criteria unconditionally. Having it optionally positionally available on the CLI is a bit weird because we already have an optional positional arg due to "0.1.0" and "0.1.0 0.2.0" both being valid inputs.

PR #86, opened 2022-05-20
Tweak suggest display

bholley:
Iterated a bit locally to make the output nicer in various ways.


bholley:
@Gankra r?

src/resolver.rs
```diff
@@ -1337,48 +1338,58 @@ impl<'a> Report<'a> {
         }
 
         suggestions.sort_by_key(|item| item.rec.diffstat.count);
+        let mut by_criteria = StableMap::new();
```
Gankra, line 1341: I don't feel like StableMap is the right one here. Presumably you want BTreeMap so that the criteria are sorted by name, and not just "whatever order they happened to be inserted in"? StableMap is mostly just used for the file formats to minimize random churn on load+save.

tests/snapshots/test_cli__test-project-suggest.snap
```diff
@@ -1,22 +1,23 @@
...
-    winapi:0.3.9        for safe-to-run     (181329 lines)  (used by atty, mio, and 9 others) 
+recommended audits for safe-to-deploy:
+    cargo vet inspect atty 0.2.14         (used by clap)                              (508 lines)
+    cargo vet inspect termcolor 1.1.3     (used by clap)                              (2585 lines)
+    cargo vet inspect os_str_bytes 6.0.0  (used by clap)                              (2918 lines)
+    cargo vet inspect textwrap 0.15.0     (used by clap)                              (6165 lines)
+    cargo vet inspect indexmap 1.8.1      (used by h2, clap)                          (9452 lines)
+    cargo vet inspect reqwest 0.11.10     (used by test-project)                      (21663 lines)
+    cargo vet inspect serde_json 1.0.79   (used by reqwest, test-project)             (22855 lines)
+    cargo vet inspect tokio 1.17.0        (used by h2, hyper, reqwest, and 4 others)  (91278 lines)
```
Gankra, line 13: Hmm so one problem with this new ordering is that the column alignment now forces the high-variance thing to the middle instead of the end. This basically forces every line to have "worst-case width" and means if a terminal is a bit too narrow then ~every line will wrap instead of only the longest.

Gankra (approved):
I like it, a couple nits.

src/resolver.rs
bholley: Good point, will fix.

tests/snapshots/test_cli__test-project-suggest.snap
bholley: I'm willing to live with this for now and see how many people complain.

PR #87, opened 2022-05-20
Certify fixups

bholley:
r? @Gankra

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -672,25 +672,20 @@ fn cmd_certify(out: &mut dyn Write, cfg: &Config, sub_args: &CertifyArgs) -> Res
     }
 
     // Print out the EULA and prompt
-    write!(
-        out,
-        "I, {}, certify that I have audited ",
-        user_info.username
-    )?;
-    match &kind {
+    let what_version = match &kind {
         AuditKind::Full { version, .. } => {
-            write!(out, "version {} ", version)?;
+            format!("version {}", version)
         }
         AuditKind::Delta { delta, .. } => {
-            write!(out, "from version {} to {} ", delta.from, delta.to)?;
+            format!("the changes from version {} to {}", delta.from, delta.to)
         }
         AuditKind::Violation { .. } => unreachable!(),
-    }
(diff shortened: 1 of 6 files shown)


src/criteria/safe-to-deploy
```diff
@@ -0,0 +1,14 @@
+This crate will not introduce a serious security vulnerability to production
```
Gankra, line 1: Can we properly mark these as .txt files?

src/main.rs
```diff
@@ -672,26 +673,21 @@ fn cmd_certify(out: &mut dyn Write, cfg: &Config, sub_args: &CertifyArgs) -> Res
...
-        "of {} in accordance with the following criteria:",
-        sub_args.package
-    )?;
-    writeln!(out, "{}\n", eula)?;
+    };
+    let statement = format!(
+        "I, {}, certify that I have audited {} of {} in accordance with the following criteria:",
+        user_info.username, what_version, sub_args.package,
+    );
+    write!(out, "\n{}\n\n", textwrap::fill(&statement, 80).yellow().bold())?;
```
Gankra: So part of the reason I imported the Term library just for one little stdin below is because it also has stuff for coloured output: https://github.com/mitsuhiko/insta/blob/master/cargo-insta/src/cli.rs#L186-L202

I just haven't had the energy to dig into the details and also want to look into what the story is for properly enabling/disabling styling based on what you're outputting to. Not sure if my `write!(out)` machinery needs to be completely replaced to handle that or what :s

Gankra (changes requested):
Hmm wish I wrote some tests for this in tests/test-cli.rs so it was easier to see the final result. The basic approach seems fine but I wish I had a better sense for what to use for terminal stuff and I feel like we should hash that out if we're doing more of it.

src/criteria/safe-to-deploy
bholley: Done.

src/main.rs
bholley: Switched over to that.

PR #89, opened 2022-05-20
Allow specifying various things as either a string or a vec

bholley:
r? @Gankra


Gankra (approved):
Neat! I was waiting to see how long it would take you to notice I had been punting on this x3

Issue #93, opened 2022-05-27
Add machine parsable output mode


glandium:
While the output of `cargo vet` could somehow be processed with regexp and whatnot, it would be even easier if there was a machine parsable output format.

Cc: @Gankra

PR #97, opened 2022-05-27
Add machine-readable output with --output-format=json

Gankra:
Machine-readable output can be set with `--output-format=json`. You can also choose to write it to a file with `--output-file=my_output.json`.

Basic checking for success/fail can be done by checking if "conclusion" == "success". Schema is currently a big *shrug*.


Gankra:
Fixes #93

PR #98, opened 2022-05-27
Document the registry

bholley:


bholley:
We probably don't need to implement this for a while, but I want to have it in the docs so that the plan is clear.

Gankra:
Oh sorry, I saw this on friday but was too fried to look at it. This is... a surprising development but it's... possible? Having it as a random static file in this repo is deeply strange though. You would think you would want to host in on an actual website/service...?

PR #121, opened 2022-06-02
Add locks and proper file management facades

### LINKED ISSUE #115: Implement proper locking of the cache
Just as cargo must acquire a global system "lockfile" before accessing .cargo, so too should we before before accessing our own caches.

I have sketched out a place to do this here:

https://github.com/bholley/cargo-vet/blob/895992c364e079e07829aedb5f959d1b55ee064a/src/main.rs#L1569-L1572

But I don't know enough about the subtleties of lockfiles (which I hear are very cursed and messy) to implement this with my current knowledge.

We may also want a way to permit a kind of... read-write locking? Perhaps with merging? To allow for multiple concurrent runs to grab the current diff-cache, compute new values, and merge them back in?

This isn't a big deal right now since in CI with --locked we want #105 so there's no need to look at the cache at all. Of course in local usage if someone invokes `vet` multiple times it can lead to corruption of the cache, which, is not great? Although this might come into play more with certain versions of #85 where we try to "remember" what the last command is to automatically choose what is being certified, and the cache is the natural place to record that fact.

Gankra:
Fixes #115 

* New FileLock type for managing a lockfile, used for both the cache and supply-chain (separate locks)
* New Store type that wraps config.toml/audits.toml/imports.lock
  * Manages acquiring the FileLock
  * Manages loading the files and validating their consistency (stubbed, not implemented)
  * Manages fetching imports and checking for criteria changes (checking is stubbed)
  * Manages storing the files back (only done when manually requested by a command to avoid changes on-error)
* All commands that touch supply-chain now unconditionally acquire it and commit it on success
  * This means every (non-freestanding, non-debug) command will try to validate+fetch-imports+fmt

Drive-by changes:

* `notes` on audits were moved back up due to [limitations in toml-rs around nested tables](https://github.com/alexcrichton/toml-rs/issues/142). More aggressively trying to fmt+commit exposed this issue
* --readonly-lockless was added as a hack to disable the supply-chain lock for concurrent tests

### DIFF
--- a/src/format.rs
+++ b/src/format.rs
@@ -112,10 +112,10 @@ pub struct CriteriaEntry {
 #[derive(serde::Serialize, serde::Deserialize, Debug, Clone)]
 pub struct AuditEntry {
     pub who: Option<String>,
+    pub notes: Option<String>,
     pub criteria: String,
     #[serde(flatten)]
     pub kind: AuditKind,
-    pub notes: Option<String>,
 }
 
 #[derive(serde::Serialize, serde::Deserialize, Debug, Clone)]
--- a/src/format.rs
+++ b/src/format.rs
@@ -9,8 +9,6 @@ use serde::{
     Deserialize, Deserializer, Serialize, Serializer,
 };
 
-use crate::VetError;
-
 pub type StableMap<K, V> = linked_hash_map::LinkedHashMap<K, V>;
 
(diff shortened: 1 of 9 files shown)


Gankra:
Oh whoops I missed some cases, one sec

Gankra:
There we go

PR #123, opened 2022-06-03
Sort the unaudited table

bholley:
At present, regenerate-unaudited will append new things to the end of the file. I think we should just always sort it.

I'm tempted to sort the other tables as well, but less certain about it.

### DIFF
--- a/src/format.rs
+++ b/src/format.rs
@@ -1,6 +1,7 @@
 //! Details of the file formats used by cargo vet
 
 use core::fmt;
+use std::collections::{BTreeMap, BTreeSet, HashMap, HashSet};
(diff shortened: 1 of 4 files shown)


bholley:
r? @Gankra 

CC @glandium

Gankra (approved):
lgtm pending clippy fixes

note that I already started establishing other typedefs for BTreeMap and HashMap in case we ever want to swap out impl details. Maybe those defs should be moved to format.rs?

https://github.com/bholley/cargo-vet/blob/37e5270fb828f4ba00c39927c94820e30883d8e5/src/resolver.rs#L16-L19

PR #126, opened 2022-06-03
Add fetch-imports and restrict it to only implicitly happening with b…

### LINKED ISSUE #122: Most command should not re-fetch imports
I think what I'd propose here is that we move fetching to a dedicated `cargo vet fetch-imports` command, and then have just bare (non-locked) `cargo-vet` implicitly fetch.

Gankra:
…are vet

Fixes #122


Gankra:
(this will merge-conflict with the other PR since they both changed the CLI snapshots)

PR #131, opened 2022-06-04
migrate logging to tracing and use panics for exits

Gankra:
This is not what I intended to be doing today but there's some connection between logging and error handling, and tracing works "better" with eyre, so this was a step1 in understanding the error-handling ecosystem better.


Gankra:
added exit-via-panics which solves the issue of having to contort around cleanup and exit

PR #133, opened 2022-06-06
Support empty lists in policy.criteria

bholley:


Gankra (approved):
looks right, though would be nice to have some more policy tests for cases you consider interesting

PR #142, opened 2022-06-07
this PR is mostly harmless

yaahc:
Only fiddled with main.rs for now, but this is a quick first pass of removing panics that didn't seem necessary. The only one I didn't end up removing was the one in `cmd_certify` which outputs 'rejected certification' via `write!` instead of `error!`.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -13,7 +13,7 @@ use std::{fs::File, io::Write, panic, path::PathBuf};
 use cargo_metadata::{Metadata, Package, Version};
 use clap::{ArgEnum, CommandFactory, Parser, Subcommand};
 use console::{style, Term};
-use eyre::Context;
+use eyre::{eyre, WrapErr};
 use flate2::read::GzDecoder;
 use format::{AuditEntry, AuditKind, Delta, DiffCache, DiffStat, MetaConfig, VersionReq};
 use reqwest::blocking as req;
@@ -841,13 +841,9 @@ fn real_main() -> Result<(), VetError> {
 
     info!("Running: {:#?}", cmd.cargo_command());
 
-    let metadata = match cmd.exec() {
-        Ok(metadata) => metadata,
-        Err(e) => {
-            error!("'cargo metadata' failed: {}", e);
-            panic_any(ExitPanic(-1));
-        }
-    };
+    let metadata = cmd
+        .exec()
+        .wrap_err("'cargo metadata' exited unsuccessfully")?;
 
     // trace!("Got Metadata! {:#?}", metadata);
     trace!("Got Metadata!");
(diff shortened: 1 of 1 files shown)


src/main.rs
```diff
@@ -841,13 +841,9 @@ fn real_main() -> Result<(), VetError> {
...
-    let metadata = match cmd.exec() {
-        Ok(metadata) => metadata,
-        Err(e) => {
-            error!("'cargo metadata' failed: {}", e);
-            panic_any(ExitPanic(-1));
-        }
-    };
+    let metadata = cmd
+        .exec()
+        .wrap_err("'cargo metadata' exited unsuccessfully")?;
```
yaahc, line 846: I noticed that you concatenated errors manually in a lot of places. Since eyre already does that automatically for source errors I opted to replace these with calls to `wrap_err` which creates a wrapping error from the given string and sets the original error as the source of that error.

src/main.rs
```diff
@@ -868,35 +864,28 @@ fn real_main() -> Result<(), VetError> {
...
+
+    // FIXME: what is `store.path` relative to here?
+    let package_metacfg = metadata
+        .root_package()
+        .and_then(|r| r.metadata.get(PACKAGE_VET_CONFIG))
+        .map(|cfg| {
+            MetaConfigInstance::deserialize(cfg)
+                .wrap_err("Root package had [{PACKAGE_VET_CONFIG}] but it was malformed")
+        })
+        .transpose()?;
```
yaahc, line 885: I changed these to use closures still initially and didn't like it personally so I pushed a second commit to change these to use method chaining, if you prefer the old it should be easy to back this out and go back to a version more consistent with what was in place before, and I can also change the return type on the closures to be just `Option<_>` instead of the full type which I wrote in in the previous commit

src/main.rs
```diff
@@ -1130,14 +1117,15 @@ fn cmd_certify(out: &mut dyn Write, cfg: &Config, sub_args: &CertifyArgs) -> Res
     let eula = if let Some(eula) = eula_for_criteria(&store.audits, &criteria) {
         eula
     } else {
-        error!("couldn't get description of criteria");
-        panic_any(ExitPanic(-1));
+        Err(eyre!("couldn't get description of criteria"))?
```
yaahc, line 1120: I'd reword this to describe the state rather than the activity that failed, though I don't have enough domain knowledge to know what wording to suggest instead. Something like "there is no description of the criteria available"

src/main.rs
Gankra: This is way nicer, the closures were me experiencing defeat at adapter-foo (TIL about transpose!).

src/main.rs
Gankra: Hmm so the purpose of the print-and-exit pattern is to not include any of the more "debug-like" features of Report like the backtrace (while still allowing it to exist for things like i/o errors).

In https://github.com/bholley/cargo-vet/pull/140/files#diff-42cb6807ad74b3e201c5a7ca98b911c5fa08380e942be6e4ac5807f8377f87fcR1190

I just changed the pattern to `writeln!(out` while still doing a PanicExit. This would make them more debug-y right? Are you planning to mess with the report format too?

yaahc: > Hmm so the purpose of the print-and-exit pattern is to not include any of the more "debug-like" features of Report like the backtrace (while still allowing it to exist for things like i/o errors).

my plan for that was to customize the error reporting to not include that extra information so we can keep the representation of errors consistent.

> I just changed the pattern to `writeln!(out` while still doing a PanicExit. This would make them more debug-y right? Are you planning to mess with the report format too?

No plans to mess with the report format, my expectation is that any errors that get emitted as part of the report will immediately be discarded and will be effectively represented further in the process lifecycle by the "report contained errors" error that gets emitted at the end.

Gankra: Hmm I'm not sure I understand the pitch here. It feels like there must be some kind of distinct error type to distinguish between user-designed errors and "errors we didn't explicitly handle and are exposed more raw"?

PR #147, opened 2022-06-08
Reconfigure CI to avoid depending on actions-rs, which our enterprise policy forbids

bholley:


bholley:
@Gankra going to just merge this to get our CI running again, but lmk if any of it seems problematic.

PR #153, opened 2022-06-09
Gardening + Breaking Things Up

Gankra:
This rebases #142 and annotates many of the fallible sites with information about the kind of error it *should* be.

There may be some output quality regressions, as e.g. a failed `vet` will now spit out

```
Vetting Failed!

1 unvetted dependencies:
  bitflags:1.3.2 missing ["safe-to-deploy"]

recommended audits for safe-to-deploy:
    cargo vet inspect bitflags 1.3.2  (used by clap, openssl, and 2 others)  (3004 lines)

estimated audit backlog: 3004 lines

Error: report contains errors

Location:
    src\main.rs:1078:13
```

------

This also splits main up into two new modules: storage and cli. Those things had grown up enough that they were getting unwieldy. Now main.rs is almost entirely just actual terminal logic. In the future I expect I will want to further split this file into an "ops" module for some high-level operations without the CLI gunk. We already have two logical ops in `init_files` and `minimize_unaudited` (because tests use them).

I wanted to get this out of the way to avoid conflicts on other people fixing things


Gankra:
fixed the vet output, let's land this

PR #161, opened 2022-06-09
Switch to cargo's infrastructure for file locking

### LINKED ISSUE #146: implement correct filelocks
I have no idea how to do these and just hacked up a simple "try to create the file, if it exists already you lose, panic". (FileLock)

Current issues:

* Panics on contention instead of blocking (ideally with a message that this is happening)
* Fails to cleanup lock if program is killed by e.g. ctrl+c
* Can't express read vs write distinctions
* Also doesn't actually have any kind of transactionality guarantees (should use tempfiles and renames more)

I *expect* that the more correct way to implement this stuff is to use APIs like flock via fs2

https://docs.rs/fs2/0.4.3/fs2/trait.FileExt.html

and to use a specific file that's "always" there (like diff-cache or config.toml) to act as the lock...?

mystor:
This required a copy of Cargo's logic for file locking to be copied in-tree. We should look into making it a seperate crate so it can be shared by other libraries in the future.

Fixes #146

### DIFF
--- a/src/storage.rs
+++ b/src/storage.rs
@@ -1,7 +1,7 @@
 use std::{
     ffi::OsString,
     fs::{self, File, OpenOptions},
-    io::{BufReader, Read, Seek, Write},
+    io::{self, BufReader, Read, Seek, Write},
     mem,
     path::{Path, PathBuf},
 };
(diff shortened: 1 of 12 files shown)


mystor:
The flock.rs file was copied with some modification from https://github.com/rust-lang/cargo/blob/4d92f07f34ba7fb7d7f207564942508f46c225d3/src/cargo/util/flock.rs

src/storage.rs
```diff
@@ -328,51 +324,44 @@ impl Cache {
...
-            if let Ok(command_history) = load_command_history(&command_history_path) {
-                command_history
-            } else {
-                // Might be our first run, create a fresh command_history that we'll write back at the end
-                CommandHistory::default()
-            };
+        let command_history = || -> Result<CommandHistory, VetError> {
+            load_json(File::open(&command_history_path)?)
+        }()
+        .unwrap_or_default();
```
Gankra, line 364: these are kinda goofy, we should just use and_then

Gankra (approved):
like the general approach

PR #163, opened 2022-06-09
Remove spelling/usage error in index.md

zaataylor:
Hi there! 👋 😄 This is a _very_ small change. I replaced "ratched down down" with "ratcheted down" to make it clearer that the size of the list of exceptions will be slowly decreased over time.


bholley:
Whoops, good catch, and thanks for the patch!

PR #166, opened 2022-06-09
Properly unwind the stack when invoking exec

bholley:


bholley:
Going to go ahead and merge this because I want to get a new rev into m-c before I head out for the day.

Issue #167, opened 2022-06-09
audit code for paths where network access can happen with --frozen


Gankra:

Gankra:
done in #173

Issue #132, opened 2022-06-06
Async processing


Gankra:
Currently the MVP was quickly hacked up with sync reqwest and no multi-threading just to keep things simple while everything is in flux. In the medium-term we should properly cleanup the code to be async and spin up multiple concurrent tasks where it makes sense.

I've already done some of that work in rust-minidump here: https://github.com/rust-minidump/rust-minidump/blob/main/breakpad-symbols/src/http.rs

Which only gets more complicated here: https://github.com/rust-minidump/rust-minidump/pull/558

Although even that is only "partially" async because the primary concurrency model for minidump-stackwalk is actually just spawning multiple instances in separate processes and them knowing how to transitionally add things to the cache. That said I have been thinking about how to make it "fully" async+multithreaded here: https://github.com/rust-minidump/rust-minidump/issues/561

I am finding myself doing a lot of duplicated logic and work between these two codebases, and it feels like there could be some very useful to factor out nugget for all the concurrent locking+caching+downloading+parsing machinery.

--------------

Things that can potentially be done concurrently in their own task:

* Each package's `suggest`
  * Want to be able to start up multiple suggests at once to queue up all the expensive subtasks below
* Each download of an imported audits.toml
  * Can we start parsing/validating it *while it's downloading*? (probably not worth it...?)
* Each fetch of a package from crates.io
  * Can we start SHAing and untarring it *while it's downloading* like we do streaming-parse in rust-minidump?
* Each diffstat of two packages (unclear if `git diff --no-index` is safe/useful to spawn multiple parallel instances of)

See https://github.com/rust-minidump/rust-minidump/issues/561 where I detail some of the machinery you would want to have to make this work.

Gankra:
If we do want to share some code with rust-minidump on this, one thing I have been thinking about a lot is how these two codebases increasingly look like standard batch compilers, and all these various caching phases increasingly look like "incremental compilation". To that end it might be worth building on a proper incremental framework like [salsa](https://github.com/salsa-rs/salsa/) (or specifically [gluon's async fork](https://github.com/Marwes/salsa/tree/gluon_salsa). 

But we're in a kind of interesting situation where we actually have very little invalidation of our caches (we tend to cache immutable facts that should never change ever). So probably we wouldn't need the core features of this... unless we realize actually we want some invalidatable caches and then it would have been nice to have this.

Gankra:
[...]

Issue #90, opened 2022-05-23
treat dependency overrides as third-party


bholley:
STR:
cd fresh-mozilla-central-clone
cargo vet init
cargo vet suggest

Resulting output: [mio.txt](https://github.com/bholley/cargo-vet/files/8758330/mio.txt)

Note that 'mio' appears several times as a reverse-dependency, but doesn't show up as a recommendation for inspect. We should figure out why that is and resolve the inconsistency.

Gankra:
Mio is being weirdly explicitly patched in here:

https://searchfox.org/mozilla-central/rev/301e9545863b14ff85aa110fabd4274da4593f9d/Cargo.toml#123

As a result in `cargo metadata` it is showing up as a path-dep instead of a crates.io dep (that is vendored). So with the input we get in `vet` has us concluding that it's a first-party crate and not eligible for auditing.

Gankra:
Here's our current first-party graph:


```mermaid
            graph LR
            node8[app_services_logger:0.1.0]
            node15[async-task:4.0.3]
            node21[autocfg:1.0.1]
            node22[baldrdash:0.1.0]
            node26[bench-collections-gtest:0.1.0]
            node35[bits:0.2.0]
            node36[bits_client:0.2.0]
            node37[bitsdownload:0.1.0]
            node42[bookmark_sync:0.1.0]
            node49[cascade_bloom_filter:0.1.0]
            node51[cert_storage:0.0.1]
            node56[chardetng:0.1.9]
            node57[chardetng_c:0.1.2]
            node72[coremidi:0.6.0]
            node92[d3d12:0.4.1]
            node99[defaultagent-static:0.1.0]
            node101[derive_common:0.0.1]
            node115[encoding_glue:0.1.0]
            node120[error-support:0.1.0]
            node133[fluent-ffi:0.1.0]
            node135[fluent-langneg-ffi:0.1.0]
            node139[fog:0.1.0]
            node140[fog-gtest:0.1.0]
            node141[fog_control:0.1.0]
            node157[gecko-profiler:0.1.0]
            node158[gecko_logger:0.1.0]
            node159[geckodriver:0.31.0]
            node160[geckoservo:0.0.1]
            node163[gkrust:0.1.0]
            node164[gkrust-gtest:0.1.0]
            node165[gkrust-shared:0.1.0]
            node166[gkrust_utils:0.1.0]
            node175[glsl-to-cxx:0.1.0]
            node177[gluesmith:0.1.0]
[...]

PR #173, opened 2022-06-10
WIP async downloads

Gankra:
This rejigs a crapton of code to make all networking completely explicit and async.

To make this more tolerable, some things that randomly do networking (notably compute_suggest) have been hoisted out of methods and must now have their result passed in, keeping those parts of the code "network pure".

To keep the rest of the code logically sync/main-threaded, we now spin-up a tokio runtime whenever we have a burst of "need to do some async work". One of the nastier consequences of this is that anything that wants to be async needs to be rooted in all `'static` inputs even though we subsequently block on them, so i had to rejig some stuff to take e.g. String instead of &str or be a free function instead of a method. Some of this could be cleaned up by requiring `Arc` instead of `&`, but I'm still feeling out the design space.

-----------

This is motivated by #90, #132, and #167.

For #132, this is only a partial answer as it only parallelizes networking in places where queries are non-conflicting "by construction" and does not allow for greater parallelism in suggest by kicking off every package's suggest in parallel. In particular a "first suggest" will still be very slow because none of the git-diffing or even cache-sniffing is parallel, only the downloading.

For #167, it now makes all code "trivially" --frozen correct, as the only way to get network access is to call something that checks that flag.

For #90, it was just going to be another situation where I need to do a bunch of network-poking and so I wanted to have a better story for when/where networking can happen. To complete it (which is what this PR will do when it's complete) I will need to further factor the code to pull DepGraph construction out of `resolve` and instead make the DepGraph an *input* to resolve. This is because computing is_third_party will become a value that depends on the Network to check if a non-crates.io dep *does* exist on crates.io.


bholley:
> This is because computing is_third_party will become a value that depends on the Network to check if a non-crates.io dep _does_ exist on crates.io.

@Gankra can't we just figure that out by examining the registry, which shouldn't require any additional fetching?

Gankra:
Possibly, it's not clear to me how "up to date" I can assume it will be / how safe it is to access without cargo / if cargo has a usable API for bulk-querying it. Also I wasn't sure if you still wanted to do fuzzy heuristics based on the delta from the "real" package.

bholley:
Plan of record from #90 is not to do the fuzzy heuristic for now.

As for the registry being up-to-date: I think that happens implicitly as soon as you query cargo metadata in the non--frozen case?

src/network.rs
```diff
@@ -0,0 +1,111 @@
...
+            .client
+            .get(url.clone())
+            .send()
+            .await
+            .and_then(|res| res.error_for_status())
+            .wrap_err_with(|| format!("Failed to download {}", url))?;
+
+        let mut partial_download = NamedTempFile::new_in(&self.in_progress_download_dir)
+            .wrap_err("could not create tempfile for download")?;
+        let mut output = vec![];
```
mystor, line 75: IIRC the only consumer right now doesn't care about this information, so it should be possible to avoid buffering this up in memory. If we want to do a sha digest to verify the file or similar, we can do that incrementally without buffering the full download in memory.

src/network.rs
```diff
@@ -0,0 +1,111 @@
...
+        while let Some(chunk) = res.chunk().await? {
+            let network_bytes = &chunk[..];
+            partial_download.write_all(network_bytes)?;
```
mystor, line 78: This call to write does blocking I/O. You could theoretically use the async I/O APIs in tokio for this, but I'm pretty sure they just spawn a background thread to do the writes, so perhaps we can just wrap this with a `tokio::task::spawn_blocking` or a `tokio::task::block_in_place`? You could also convert the tempfile into a `tokio::fs::File` with `tokio::fs::File::from_std(partial_download.into_file())` and then do async writes to the file through that API.

src/resolver.rs
```diff
@@ -2435,7 +2447,8 @@ impl FailForVet {
         &self,
         out: &mut dyn Write,
         report: &ResolveReport,
-        cfg: &Config,
+        _cfg: &Config,
```
mystor, line 2450: Perhaps we should drop this argument now?

Cargo.toml
```diff
@@ -26,12 +26,13 @@ flate2 = { version = "1.0.3", default-features = false, features = ["zlib"] }
 home = "0.5.3"
 libc = "0.2"
 nom = "7.1.1"
-reqwest = { version = "0.11.10", features = ["blocking"] }
+reqwest = { version = "0.11.10", features = ["rustls-tls"] }
```
mystor, line 29: Why enable `rustls-tls`? IIRC the native TLS will actually still exist as a dependency and  it pulls in quite a few extra dependencies.

mystor (commented):
I like the new Network design quite a bit, it's nice to have that explicit for the purposes of making sure we bubble that information around. It does look like the mixture of async and sync I/O will be a bit awkward going forward unfortunately, as we do so much local I/O, but overall I think this is a solid improvement.

The main things I think I'd like to see tweaked are dropping some of the extra unnecessary new dependencies if we can, and perhaps spawning a single runtime in main to avoid all of the duplicated runtime spawning code.

src/main.rs
```diff
@@ -1228,18 +1258,23 @@ fn cmd_fetch_imports(
...
+        }
     }
-    store.commit()?;
+
+    // ERRORS: just a warning that you're holding it wrong, unclear if immediate or buffered,
+    // or if this should be a hard error, or if we should ignore the --locked flag and
+    // just do it anyway
+    writeln!(
+        out,
+        "warning: ran fetch-imports with --locked, this won't do anything!"
```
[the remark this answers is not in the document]
Gankra: --frozen implies --locked, so this is "correct" but I agree it might be worth being explicit about...

src/network.rs
```diff
@@ -0,0 +1,111 @@
...
+
+use crate::{PartialConfig, VetError};
+
+pub struct Network {
+    /// The HTTP client all requests go through
+    client: Client,
+    /// A temp dir where partial downloads for things that want to be persisted
+    /// to disk will be built up before being transactionally put in their final
+    /// destination with a rename.
+    in_progress_download_dir: PathBuf,
```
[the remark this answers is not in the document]
Gankra: That explains this old "same filesystem" stuff that minidump-stackwalk has for this exact pattern :(

https://github.com/rust-minidump/rust-minidump/tree/main/minidump-stackwalk#--symbols-cache-symbols_cache

src/resolver.rs
Gankra: I want to keep some of these spiderings around in case we ever have more formatting flags we want to dump into here.

src/network.rs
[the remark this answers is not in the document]
mystor: Yeah, IIRC it's often an issue when using the temporary directory for the tmp file, as IIRC some platforms store that in a ramdisk so it's different filesystem than anywhere where you'd actually persist the file.

src/main.rs
```diff
@@ -1551,25 +1590,42 @@ fn eula_for_criteria(audits: &AuditsFile, criteria: CriteriaStr) -> Option<Strin
...
+        criteria
+    );
+
+    // Now try the description
+    if let Some(eula) = criteria_entry.description.clone() {
+        return eula;
+    }
+
+    // If we get here then there must be a URL, try to fetch it. If it fails, just print the URL
+    let url = Url::parse(criteria_entry.description_url.as_ref().unwrap()).unwrap();
```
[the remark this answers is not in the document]
Gankra: this should be handled in Store::validate (but it's not right now)

Cargo.toml
mystor: we might need a `no-default-features = true` flag passed in then to make sure to not build the openssl backend in addition to rustls-tls.

Gankra: Ah you are correct https://github.com/rust-minidump/rust-minidump/blob/main/breakpad-symbols/Cargo.toml#L30

Gankra:
This PR is an absolute mess, but:

* Everything is implemented except for some tokio quality things
  * ❌ I still make fresh runtimes all over the place
  * ❌ I don't mark certain things as Truly Blocking
* This implements #90
  * We now use crates-index and have pretty spiffy access for reading the crates.io registry index
  * Registry access code has now been cleaned up and simplified because that library handles discovery
* This fixes up the huge pile of clap deprecations 

I'm not terribly proud of this but the PR is getting unwieldy and generally moving things in the right direction, so I think it might be a good idea to just land it and iterate on it in followups?

src/main.rs
```diff
@@ -793,18 +801,27 @@ fn cmd_certify(out: &mut dyn Write, cfg: &Config, sub_args: &CertifyArgs) -> Res
...
+    if !sub_args.accept_all {
+        for criteria in &criteria_names {
+            let handle = runtime.spawn(eula_for_criteria(
+                network.clone(),
+                store.audits.criteria.clone(),
+                criteria.to_string(),
+            ));
+            eulas.insert(*criteria, handle);
+        }
+    }
```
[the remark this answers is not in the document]
Gankra: *shrug* it happens, worth the extra pair of hands

src/main.rs
```diff
@@ -1197,15 +1223,58 @@ fn cmd_vet(out: &mut dyn Write, cfg: &Config) -> Result<(), VetError> {
     trace!("vetting...");
 
     let mut store = Store::acquire(cfg)?;
+    let network = Network::acquire(cfg);
+
     if !cfg.cli.locked {
```
mystor, line 1228: We may also want a flag to override this without specifying locked at some point.

Cargo.lock
```diff
@@ -202,21 +207,25 @@ dependencies = [
...
+checksum = "2519c91ad7a6e3250a64fb71162d2db1afe7bcf826a465f84d2052fd69639b7a"
 dependencies = [
- "core-foundation-sys",
- "libc",
+ "git2",
+ "hex",
+ "home",
+ "memchr",
+ "num_cpus",
+ "rayon",
```
mystor, line 220: Let's turn off the `"parallel"` feature of `crates-index` to avoid pulling in this rayon dependency

mystor (approved):
Let's land this once we get rid of the rayon dependency. There's lots of cleanup here but we can get to that after it's landed.

Gankra:
landing as-is, let's iterate on this in follow-ups. the audit-as-crates-io UX needs to be iterated on in particular

PR #181, opened 2022-06-14
Wrap long inline arrays and tables over multiple lines

### LINKED ISSUE #176: line-wrap long items?
Notable candidates:

* dependency-criteria (inline array gets crammed all on one line)
* notes (currently no way to input multiple lines in interactive prompt, creates very long lines)

mystor:
This should help improve the readability of very complex toml files by wrapping complex fields over multiple lines (either as a non-inline table or as a multi-line array literal) when they get too long.

The current thresholds selected are completely arbitrary and can be changed in the future.

Fixes #176

### DIFF
--- a/src/serialization.rs
+++ b/src/serialization.rs
@@ -110,10 +110,38 @@ pub mod dependency_criteria {
     }
 }
 
-/// tables with these names will be rendered by TomlFormatter as inline tables,
-/// rather than dotted tables.
-fn always_render_inline(key: &str) -> bool {
-    key == "dependency-criteria"
+/// Inline arrays which have a representation longer than this will be rendered
+/// over multiple lines.
+const ARRAY_WRAP_THRESHOLD: usize = 80;
+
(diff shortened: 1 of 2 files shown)


src/serialization.rs
```diff
@@ -110,10 +110,38 @@ pub mod dependency_criteria {
...
+    // serialized representation of the value in the TOML document, which we can
+    // use to determine the line length in utf-8 characters.
+    let value_repr = value.to_string();
+    key.chars().count() + value_repr.chars().count() + DECORATION_LENGTH
+}
+
+/// Tables with these names will be rendered by TomlFormatter as inline tables,
+/// rather than dotted tables unless their inline representation exceeds
+/// `INLINE_TABLE_THRESHOLD` UTF-8 characters in length.
+fn render_table_inline(key: &str, value: &toml_edit::Item) -> bool {
```
Gankra, line 140: i'd prefer if this was worded more as a question, like `table_should_be_inline`. As is it sounds like it has a side-effect.

Gankra (approved):
nice

PR #185, opened 2022-06-14
Use the 'dirs' crate to store our cache in a platform-appropriate cache directory

### LINKED ISSUE #184: Consider storing the cache in $XDG_CACHE_HOME on platforms where that's applicable
Currently the cache is stored in the system's temporary directory. On platforms where XDG is relevant, we could consider using `$XDG_CACHE_HOME` instead. 

https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html
>  `$XDG_CACHE_HOME` defines the base directory relative to which user-specific non-essential data files should be stored. If `$XDG_CACHE_HOME` is either not set or empty, a default equal to `$HOME/.cache` should be used.

mystor:
This seemed straightforward enough to do, so I figured I'd post a patch for it based on @glandium's suggestion, so it's ready if we decide to go with it.

Fixes #184

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -85,6 +85,7 @@ dependencies = [
  "clap-cargo",
  "console",
  "crates-index",
+ "dirs",
  "eyre",
  "flate2",
  "home",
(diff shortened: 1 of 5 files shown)


Gankra:
Are these dirs candidates for gc..? Everything just says "oh it's the cache" and idk what that implies.

mystor:
To my understanding that is a OS-specific-behaviour question. I'm not sure by default when places like `XDG_CACHE_DIR` are cleared out.

IIRC on some systems `/tmp` is a ramdisk, but I expect `$XDG_CACHE_DIR` is generally not by default.

bholley:
Closing this in favor of #191.

PR #194, opened 2022-06-15
Improve the output of guess-deeper

mystor:
This makes a few changes to resolution to track our confidence in the
requirements for a specific dependency, and propagate that through to
emitting suggestions and vet error reports.

Lines corresponding to uncertain requirements will be dimmed when
written to a terminal, and will be ordered after confident lines when
listing.

This mode will be made the default in a follow-up.


src/main.rs
```diff
@@ -1176,7 +1176,7 @@ pub fn minimize_unaudited(
             for item in new_items {
                 for criteria in report
                     .criteria_mapper
-                    .criteria_names(&item.suggested_criteria)
+                    .all_criteria_names(&item.suggested_criteria)
```
Gankra, line 1179: this is gonna conflict with my changes but that's fine

Gankra (approved):
looks great!

PR #196, opened 2022-06-15
criteria blitz

### LINKED ISSUE #138: implement violation criteria-masking
Currently violations have criteria but we don't look at them at all. I believe the intended semantic is that the violation says "any attempt to require this criteria of these versions is an error". So right now we're ~basically interpreting everything as violation: safe-to-run. If someone has violation: safe-to-deploy it should still be legal for audits/unaudited to claim that version is safe-to-run and for you to have that package as long as only safe-to-run is required.

The mechanical interpretation of this is that if we have "violation X" then any criteria which implies X (including X itself) is invalid. Thankfully I believe it is impossible to conclude "my-third-party-package: X" without an audit/unaudited entry that explicitly asserts "my-third-party-package: criteria-that-implies-X", so we never have to get violation checking mixed in with policies or path-search. We can do the same thing we're doing now where we just check the audits/unaudited during pre-processing and spit out an error if there's a conflict.

### LINKED ISSUE #189: Implement criteria "decomposition"
1. unaudited/audits should take a list of criteria instead of a single criteria
2. implicit dependency-criteria (including dependence on previous versions of yourself in deltas) should implicitly reduce to the current criteria

Because we already evaluate every package for every criteria individually, we already know what criteria we're trying to "prove". When validating an edge in the audit-graph we should not make the implicit dependency-criteria be "the criteria on the audit" but "the single criteria we are currently trying to evaluate".

The upshot of this is that if `glean` is audited for safe-to-deploy but `glean-core` is only safe-to-run, we will successfully audit `glean` for safe-to-run instead of just completely discarding it. Similarly if you are audited for `[x, y]` and your dependency only has `x`, then you should be certified for `x`.

*explicit* dependency-criteria gets no such treatment and should maintain its current un-decomposable nature.

Gankra:
fixes #138 
fixes #189


src/resolver.rs
```diff
@@ -1,3 +1,63 @@
+//! The Resolver is the heart of cargo-vet, and does all the work to validate the audits
```
mystor, line 1: Yay more resolver documentation! This is nice :-)

src/resolver.rs
```diff
@@ -940,7 +1028,7 @@ pub fn resolve<'a>(
     store: &'a Store,
     guess_deeper: bool,
 ) -> ResolveReport<'a> {
-    let _resolve_span = trace_span!("resolve").entered();
+    let _resolve_span = trace_span!("validate").entered();
```
mystor, line 1031: Why this name change?

src/resolver.rs
```diff
@@ -1279,17 +1364,68 @@ fn resolve_third_party<'a>(
...
+        // The first 3 cases are correctly handled by audit.contains(violation)
+        // but the last one isn't. I think the correct solution to this is to
+        // *for each individual entry in the violation* do audit.contains(violation).
+        // If any of those queries trips, then it's an ERROR.
+        //
+        // Note that this would also more correctly handle [safe-to-deploy, safe-to-run]
+        // as a violation entry because it would effectively become safe-to-run instead
+        // of safe-to-deploy, which is the correct and desirable behaviour!
+        //
+        // So here we make a criteria set for each entry in the violation.
```
mystor, line 1386: Had to read this 3 times, but I'm pretty sure you're right and this should work.

src/resolver.rs
```diff
@@ -2526,6 +2689,18 @@ impl FailForViolationConflict {
...
         }
 
+        fn print_unaudited_entry(
+            out: &mut dyn Write,
+            entry: &UnauditedDependency,
+        ) -> Result<(), VetError> {
+            writeln!(out, "unaudited {}", entry.version)?;
+            writeln!(out, "      criteria: {:?}", entry.criteria)?;
+            if let Some(notes) = &entry.notes {
+                writeln!(out, "      notes: {notes}")?;
```
mystor, line 2699: I wonder a bit if we'd want to put a line break before the notes if they're multiline notes to avoid breaking formatting. Probably a problem for a follow-up.

mystor (approved):
This looks great, and the documentation is super clear. Thanks!

src/resolver.rs
Gankra: The diagnostic stuff I'm messing with would ideally end up with us producing a Span on the original toml and just printing it out

PR #198, opened 2022-06-16
Change the default behaviour to guess-deeper

### LINKED ISSUE #190: make --guess-deeper the default for both `vet` and `vet suggest`
UX TBD, but we want to emphasize that the top-level failures are "more important" or "better" and that the sub-failures may be irrelevant if you handle them.

mystor:
This introduces a new flag (--shallow) which can be used to restore the
previous default behaviour for suggest and vet.

Fixes #190


Gankra (approved):
I think we possibly want to invert the polarity on all the unit-tests, but I think it's good to leave that to a follow-up where we can think about that in a vacuum.

PR #199, opened 2022-06-16
Make suggest much more async, move cache to a platform-specific dir, and add a "vet gc" command

### LINKED ISSUE #132: Async processing

### LINKED ISSUE #191: Store cargo vet data in the OS cache directory and GC it
Supersedes #184.

My proposal here is:
(1) Stash everything in the dir crate's `cache_dir()`
(2) Have suggest store the last time a crate was used (directly, or implicitly via the diffcache).
(3) Implement a gc algorithm, which removes anything not used in the last month.
(4) Have any command that flushes to disk do an implicit gc.

mystor:
This is unfortunately a combination of 3 seperate changes which all ended up being interdependent. Each of the changes are described below:

<details><summary>Use the 'dirs' crate to store our cache in a platform-appropriate cache directory</summary>

This is a more persistent place to write the cache than in the temporary
directory, and we'll implement a GC mechanism on top of it to clean out
unnecessary cache files over time in a later commit.

This also renames some variables to reflect that the directory is now
explicitly a cache, rather a temporary directory.
</details>

<details><summary>More aggressively make parallelizable io async</summary>

This makes all io around networking and diffing crates async, and attempts to
run things concurrently as much as possible. For example, when generating
diffstats for suggest, all crate downloads and their diffs are run concurrently
with limits to avoid too many connections to crates.io at once.

This refactor also includes some extra logic which was added to the package
fetching logic to ensure that we update the mtime on the .crate file whenever
it is accessed, so that information can be used in the next patch for GC.

To make this change, the cache needed to be made shareable across multiple
concurrent tasks. The core mutable state was moved behind a mutex, and new
tables of async OnceCells were added to ensure that only one task would attempt
to download or diffstat each crate at a time.

With these changes, the time to perform a `cargo vet suggest` on
mozilla-central with an empty cache and use of the cargo registry cache
disabled, reduced from ~18s to ~1.5s.
</details>

<details><summary>Add a 'vet gc' command to remove old packages from the cache</summary>

This command will delete any unrecognized files in the cache directory, as well
as old/out-of-date packages. In addition, a `--clean` flag was added which is
even more aggressive, removing even the diffcache and command-history.json.
</details>

Fixes #191
Fixes #132

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -184,6 +184,15 @@ pub enum Commands {
     #[clap(disable_version_flag = true)]
     #[clap(hide = true)]
     HelpMarkdown(HelpMarkdownArgs),
+
+    /// Clean up old packages from the vet cache
+    ///
+    /// Removes  packages which haven't been accessed in a while, and deletes
+    /// any extra files which aren't recognized by cargo-vet.
+    ///
+    /// In the future, many cargo-vet subcommands will implicitly do this.
+    #[clap(disable_version_flag = true)]
+    Gc(GcArgs),
 }
 
 #[derive(clap::Args)]
@@ -323,6 +332,20 @@ pub struct AcceptCriteriaChangeArgs {}
 
 #[derive(clap::Args)]
 pub struct HelpMarkdownArgs {}
+
+#[derive(clap::Args)]
+pub struct GcArgs {
+    /// Packages in the vet cache which haven't been used for this many days
+    /// will be removed.
+    #[clap(long, action)]
+    #[clap(default_value_t = 30.0)]
+    pub max_package_age_days: f64,
+
+    /// Remove the entire cache directory, forcing it to be regenerated next
+    /// time you use cargo vet.
+    #[clap(long, action)]
+    pub clean: bool,
+}
 
 /// Cargo feature flags, copied from clap_cargo to change defaults.
 #[derive(Default, Clone, Debug, PartialEq, Eq, clap::Args)]
--- a/src/main.rs
+++ b/src/main.rs
@@ -4,6 +4,7 @@ use std::ops::Deref;
 use std::panic::panic_any;
 use std::path::Path;
 use std::sync::Arc;
+use std::time::Duration;
 use std::{fs::File, io::Write, panic, path::PathBuf};
 
 use cargo_metadata::{Metadata, Package};
@@ -86,12 +87,14 @@ impl PackageExt for Package {
     }
 }
 
-static CACHE_DIR_SUFFIX: &str = "cargo-vet";
-static CARGO_ENV: &str = "CARGO";
+const CACHE_DIR_SUFFIX: &str = "cargo-vet";
+const CARGO_ENV: &str = "CARGO";
 // package.metadata.vet
-static PACKAGE_VET_CONFIG: &str = "vet";
+const PACKAGE_VET_CONFIG: &str = "vet";
 // workspace.metadata.vet
-static WORKSPACE_VET_CONFIG: &str = "vet";
+const WORKSPACE_VET_CONFIG: &str = "vet";
+
+const DURATION_DAY: Duration = Duration::from_secs(60 * 60 * 24);
 
 /// Trick to let us std::process::exit while still cleaning up
 /// by panicking with this type instead of a string.
@@ -221,6 +224,7 @@ fn real_main() -> Result<(), VetError> {
         Some(Inspect(sub_args)) => return cmd_inspect(out, &partial_cfg, sub_args),
         Some(Diff(sub_args)) => return cmd_diff(out, &partial_cfg, sub_args),
         Some(HelpMarkdown(sub_args)) => return cmd_help_md(out, &partial_cfg, sub_args),
+        Some(Gc(sub_args)) => return cmd_gc(out, &partial_cfg, sub_args),
         _ => {
             // Not a freestanding command, time to do full parsing and setup
         }
@@ -1481,6 +1485,30 @@ fn cmd_help_md(
         is_full_command = false;
     }
 
+    Ok(())
+}
+
+fn cmd_gc(out: &mut dyn Write, cfg: &PartialConfig, sub_args: &GcArgs) -> Result<(), VetError> {
+    let cache = Cache::acquire(cfg)?;
+
+    if sub_args.clean {
+        writeln!(
+            out,
+            "cleaning entire contents of cache directory: {}",
+            cfg.cache_dir.display()
+        )?;
+        cache.clean_sync()?;
+        return Ok(());
+    }
+
+    if sub_args.max_package_age_days.is_nan() {
+        return Err(eyre!("max package age cannot be NaN"));
+    }
+    if sub_args.max_package_age_days < 0.0 {
+        return Err(eyre!("max package age cannot be negative"));
+    }
+
+    cache.gc_sync(DURATION_DAY.mul_f64(sub_args.max_package_age_days));
     Ok(())
 }
 
--- a/src/main.rs
+++ b/src/main.rs
@@ -55,8 +55,8 @@ pub struct Config {
 pub struct PartialConfig {
     /// Details of the CLI invocation (args)
     pub cli: Cli,
-    /// Path to the global tmp we're using
-    pub tmp: PathBuf,
+    /// Path to the cache directory we're using
+    pub cache_dir: PathBuf,
     /// Whether we should mock the global cache (for unit testing)
     pub mock_cache: bool,
 }
@@ -86,7 +86,7 @@ impl PackageExt for Package {
     }
 }
 
-static TEMP_DIR_SUFFIX: &str = "cargo-vet-checkout";
+static CACHE_DIR_SUFFIX: &str = "cargo-vet";
 static CARGO_ENV: &str = "CARGO";
 // package.metadata.vet
 static PACKAGE_VET_CONFIG: &str = "vet";
@@ -208,10 +208,12 @@ fn real_main() -> Result<(), VetError> {
     ////////////////////////////////////////////////////
 
     // TODO: make this configurable
-    let tmp = std::env::temp_dir().join(TEMP_DIR_SUFFIX);
+    let cache_dir = dirs::cache_dir()
+        .unwrap_or_else(std::env::temp_dir)
+        .join(CACHE_DIR_SUFFIX);
     let partial_cfg = PartialConfig {
         cli,
-        tmp,
+        cache_dir,
         mock_cache: false,
     };
 
--- a/src/main.rs
+++ b/src/main.rs
@@ -3,7 +3,6 @@ use std::mem;
 use std::ops::Deref;
 use std::panic::panic_any;
 use std::path::Path;
-use std::process::Command;
 use std::sync::Arc;
 use std::{fs::File, io::Write, panic, path::PathBuf};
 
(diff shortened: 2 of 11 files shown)


src/cli.rs
```diff
@@ -175,6 +175,15 @@ pub enum Commands {
...
     HelpMarkdown(HelpMarkdownArgs),
+
+    /// Clean up old packages from the vet cache
+    ///
+    /// Removes  packages which haven't been accessed in a while, and deletes
+    /// any extra files which aren't recognized by cargo-vet.
+    ///
+    /// In the future, many cargo-vet subcommands will implicitly do this.
+    #[clap(disable_version_flag = true)]
+    Gc(GcArgs),
```
Gankra, line 186: `#[clap(hide = true)]` is a thing you can use if you want to literally hide this from help messages, btw (I should figure out if we can get clap to group things so we can have all these debugging commands sectioned off clearly...)

src/resolver.rs
```diff
@@ -2315,129 +2316,156 @@ impl<'a> ResolveReport<'a> {
...
-                    reverse_deps.push(format!("and {} others", remainder));
-                }
-                reverse_deps.join(", ")
-            };
+        let cache = Cache::acquire(cfg)?;
+
+        let mut suggestions = tokio::runtime::Handle::current()
+            .block_on(join_all(fail.failures.iter().map(
+                |(failure_idx, audit_failure)| async {
+                    let failure_idx = *failure_idx;
```
Gankra, line 2324:
```suggestion
                |(&failure_idx, audit_failure)| async {
```

no?

src/tests.rs
```diff
@@ -897,6 +902,7 @@ fn _init_trace_logger() {
 fn mock_simple_init() {
     // (Pass) Should look the same as a fresh 'vet init'.
 
+    let _enter = TEST_RUNTIME.enter();
```
Gankra, line 905: this is a bit of a nightmare but *shrug*

src/network.rs
```diff
@@ -44,52 +43,61 @@ impl Network {
...
-            .client
-            .get(url.clone())
-            .send()
-            .await
-            .and_then(|res| res.error_for_status())
-            .wrap_err_with(|| format!("Failed to download {}", url))?;
+    pub async fn download_and_persist(&self, url: Url, persist_to: &Path) -> Result<(), VetError> {
+        let download_tmp_path = OsString::from_iter([persist_to.as_os_str(), OsStr::new(".part")]);
+        {
+            let _permit = self.connection_semaphore.acquire().await?;
```
Gankra, line 61: interesting, we want to limit network ops but not disk ops? (genuine question, curious)

src/storage.rs
```diff
@@ -419,200 +450,324 @@ impl Cache {
...
-                    let crate_name = format!("{}.crate", dir_name);
-                    let fetched_package = package_cache_dir.join(crate_name);
-                    to_download.push((name, version, fetched_package, fetched_src.clone()));
-                }
-
-                // Either this path exists or we'll download it, either way, it's right
-                fetched_src
-            };
+    #[tracing::instrument(skip(self, network))]
+    pub async fn fetch_package(
```
Gankra, line 454: there's a strange irony in async making it *better* for concurrency to define things as individual ops instead of bulk ops. I was trying to be responsible! Ah well, at least it made it easy to port the callers to joins.

src/storage.rs
```diff
@@ -419,200 +450,324 @@ impl Cache {
...
+                })
+                .await?;
+
+                // If the file isn't in our local cache, make sure to download it.
+                let file = match cached_file {
+                    Ok(file) => file,
+                    Err(_) => {
+                        let network = network
+                            .as_ref()
+                            .expect("running as --frozen but needed fetches!");
```
Gankra, line 522: Hmm now that we're doing things at this granularity we can arguably just fail this one download and any diffs that want to key off of it by storing a Result in the Once (not for this PR, followup).

Gankra: Or wait is OnceCell already doing this for us?

src/storage.rs
```diff
@@ -419,200 +450,324 @@ impl Cache {
...
+                            package,
+                            version,
+                            url,
+                            fetched_package.display()
+                        );
+                        network.download_and_persist(url, &fetched_package).await?;
+
+                        // Check if the resource is already available in our local cache.
+                        let fetched_package_ = fetched_package.clone();
+                        tokio::task::spawn_blocking(move || File::open(&fetched_package_)).await??
```
Gankra, line 539: wait dafuq? We download and then check if already available? Is this comment just wrong?

src/storage.rs
```diff
@@ -419,200 +450,324 @@ impl Cache {
...
+                    info!(
+                        "unpacking package {}:{} from {} to {}",
+                        package,
+                        version,
+                        fetched_package.display(),
+                        fetched_src.display()
+                    );
+                    // The task needs to be unpacked, so do so.
+                    tokio::task::spawn_blocking(move || {
+                        unpack_package(&file, &fetched_src)
```
Gankra, line 557: comment seems wonky

src/storage.rs
```diff
@@ -419,200 +450,324 @@ impl Cache {
...
-            let cached = self
-                .diff_cache
-                .get(package)
-                .and_then(|cache| cache.get(delta))
-                .cloned();
+            guard
+                .diffed
+                .entry((package.to_owned(), delta.clone()))
+                .or_default()
+                .clone()
```
Gankra, line 613: Wait I only just now realized you're taking a Mutex to the state and then cloning the interior once_cell out... ohhh it's an Arc ok ok. This might merit a comment at the two places you do this because it seems super wrong in a vacuum.

src/storage.rs
```diff
@@ -419,200 +450,324 @@ impl Cache {
...
+    /// Remove all files located in the `cargo-vet/empty` directory, as it
+    /// should be empty.
+    async fn gc_empty(&self) -> Result<(), VetError> {
+        let empty = self.root.as_ref().unwrap().join(CACHE_EMPTY_PACKAGE);
+        let mut empty_entries = tokio::fs::read_dir(&empty).await?;
+        while let Some(entry) = empty_entries.next_entry().await? {
+            remove_dir_entry(&entry).await?;
+        }
+        Ok(())
+    }
```
Gankra, line 695: actually now that I think about it empty/ should contain a .cargo-ok, but not a problem for this PR.

Gankra (approved):
This all looks good, nice!

src/main.rs
```diff
@@ -86,12 +87,14 @@ impl PackageExt for Package {
...
-static TEMP_DIR_SUFFIX: &str = "cargo-vet-checkout";
-static CARGO_ENV: &str = "CARGO";
+const CACHE_DIR_SUFFIX: &str = "cargo-vet";
+const CARGO_ENV: &str = "CARGO";
 // package.metadata.vet
-static PACKAGE_VET_CONFIG: &str = "vet";
+const PACKAGE_VET_CONFIG: &str = "vet";
 // workspace.metadata.vet
-static WORKSPACE_VET_CONFIG: &str = "vet";
+const WORKSPACE_VET_CONFIG: &str = "vet";
```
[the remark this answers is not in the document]
mystor: Moreso I was writing a `const` and it bothered me that the others were `static` nearby, so I changed them to be consistent

src/resolver.rs
mystor: That unfortunately doesn't work. The async closure apparently ends up capturing a reference with the lifetime of the argument to the function (i.e. `&&'a usize` rather than `&'a usize`) in that situation, and you get a borrowck error. I think it has something to do with `async` captures and patterns.

You can see a super minimal example with this: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=ed9f8169788bc5de74206bef2907450b

src/tests.rs
mystor: Yeah, I couldn't come up with a better option, so I just did a sed pass basically. I think we might want to have some macro or something like that which we just invoke at the start of a function to do setup perhaps? Would also let us set up tracing in tests for example.

src/network.rs
mystor: We need to limit network operations because otherwise we'll get network errors because crates.io will stop accepting new network connections from us. I could actually run into this on mozilla-central when fetching crates.

Technically I think we might need to do even better than this (there's an official crawler policy for crates.io, but I don't think cargo-vet will do bad enough to run into those hard rate limits: https://crates.io/policies#crawlers)

I didn't bother with the file I/O. That's already somewhat limited by the number of blocking threads in the tokio runtime (the default appears to be 512: https://docs.rs/tokio/latest/tokio/runtime/struct.Builder.html#method.max_blocking_threads), and it doesn't seem to be causing us any issues. We're limited enough by the network that file I/O being a problem doesn't come up, and we want to try to go as fast as we can with that bottleneck I'd think.

If in the future it comes up as an issue, we can look into a higher limit for other types of I/O.

src/storage.rs
mystor: because of how I'm initializing it (with the `get_or_try_init` method), if the `OnceCell` fails we'll just not initialize it, return an error, and try again next time we request it. I could also just store the result in the `OnceCell`, but that's a bit trickier, as `eyre::Report` isn't `Clone`, so this seemed like an easier approach.

src/storage.rs
mystor: Yup, the comment is just wrong, I accidentally copied it.

src/storage.rs
mystor: Yeah, I think I'm going to refactor this into a helper type like `CacheMap<K, V>` at some point, which internally is a `Mutex<HashMap<K, Arc<OnceCell<V>>>>`.

src/tests.rs
Gankra: there's already a "setup tracing" function in here, I just add it to any test I want to debug at the time

src/resolver.rs
Gankra: bizarre, ok

src/storage.rs
Gankra: hmm ok, we'll get this in followups if/when we stop using Report so arbitarily and have our own Error types.

src/cli.rs
mystor: I wasn't sure that we necessarily wanted to hide it, I could see someone wanting to run it with a specific time or with --clean if they noticed the cargo-vet cache file getting a bit big.

src/tests.rs
mystor: That's fair. If our entire stack was async from the root, we could make these all `#[tokio::test]`, but you can't `block_on` inside of a `#[tokio::test]`.

PR #206, opened 2022-06-17
Validate checksums for downloaded packages

### LINKED ISSUE #116: Implement .crate SHA validation
Whenever we have to fetch a package into our own cache, we're downloading a .crate file that has a known SHA in the registry and potentially also the lockfile. We should be responsible and compute+check that SHA.

The precise code where cargo does this:

https://github.com/rust-lang/cargo/blob/13662250abdd813755263e0de5623bd314a443db/src/cargo/sources/registry/download.rs#L85-L89

Which is using https://docs.rs/cargo-util/0.1.2/cargo_util/struct.Sha256.html

mystor:
Fixes #116


mystor:
This is just a sketch of some stuff I was poking at for doing checksum validation while downloading packages.

Issue #193, opened 2022-06-15
Move audit-as-crates-io to the policy table and implement `cargo vet regenerate`


bholley:
At the moment, `audit-as-crates-io` is its own separate table in `config.toml`, and the workflow involves copy-pasting a suggested version of that table into the file. It would be semantically cleaner to have this information (which pertains to nominally-first-party crates) in the policy table. This makes the copy-paste suggestion less viable, but it would actually be nicer anyway for the tool to take care of changing `config.toml`.

This leaves us with three "fixup" commands (regenerate-unaudited, accept-criteria-change, and this new thing). I think we should merge all of these into a single top-level `cargo vet regenerate` command with subcommands for each (`cargo vet regenerate unaudited`, `cargo vet regenerate imports`, and `cargo vet regenerate audit-as-crates-io`).

Marking this as high-priority because I would rather not check the current `audit-as-crates-io` table into m-c, which means this change blocks the next rev-bump.

Gankra:
Some loose thoughts on audit-as-crates-io I need to braindump somewhere:

Currently it very aggressively transforms the crate into "completely third-party" for anything that asks. Arguably, this means the package *shouldn't have a policy anymore* because it won't be fed through resolve_first_party... except it does still have dev-deps, and should still be processed by resolve_dev.

Still, `policy.criteria` is nonsensical in this state and `policy.dependency_criteria` is only half-enabled in this state.

This is a new unique scenario -- a third-party with dev-deps. This is actually *asserted against* in resolve_third_party, so some code is *definitely* wrong and needs to be adjusted. I'm like 70% sure that we mishandle this case in DepGraph::new itself, although I forget why. I just remember noting that in the back of my head when I implemented resolver v3.

Thankfully we do still have the distinciton of is_workspace_member separate from third_party so most of the code will organically handle this situation correctly because dev-deps are mostly determined by is_workspace_member and not is_third_party.

I suppose it's an open question if workspace-members should be allowed for audit-as-crates-io. I am inclined to say yes if-only because what exactly "is" or "isn't" a workspace-member is mildly arbitrary when you do a lot vendoring in the ways mozilla does. Maybe I'm wrong though.

We definitely need tests for audit-as-crates-io, it should not be deployed until we have some.

Gankra:
[...]

PR #207, opened 2022-06-18
audit-as-crates-io v2

Gankra:
* Makes it a part of policy
* Makes it an error to have it '= true' when the version isn't in the registry (false is fine and unchecked)
* Makes code correctly use normal_and_build_deps instead of all_deps where appropriate (handles workspace members-as-third-party)
* Removes a completely unused field from PackageNode that was an artifact of a previous refactor

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -10,7 +10,7 @@ use cargo_metadata::{Metadata, Package};
 use clap::{CommandFactory, Parser};
 use console::{style, Term};
 use eyre::{eyre, WrapErr};
-use format::{CriteriaName, CriteriaStr, PackageName};
+use format::{CriteriaName, CriteriaStr, PackageName, PolicyEntry};
 use futures_util::future::join_all;
 use network::Network;
 use reqwest::Url;
@@ -70,12 +70,15 @@ impl Deref for Config {
 }
 
 pub trait PackageExt {
-    fn is_third_party(&self, audit_as_crates_io: &SortedMap<PackageName, bool>) -> bool;
+    fn is_third_party(&self, policy: &SortedMap<PackageName, PolicyEntry>) -> bool;
 }
 
 impl PackageExt for Package {
-    fn is_third_party(&self, audit_as_crates_io: &SortedMap<PackageName, bool>) -> bool {
-        let forced_third_party = *audit_as_crates_io.get(&self.name).unwrap_or(&false);
+    fn is_third_party(&self, policy: &SortedMap<PackageName, PolicyEntry>) -> bool {
+        let forced_third_party = policy
+            .get(&self.name)
+            .and_then(|policy| policy.audit_as_crates_io)
+            .unwrap_or(false);
         let is_crates_io = self
             .source
             .as_ref()
@@ -451,7 +454,6 @@ pub fn init_files(
             imports: SortedMap::new(),
             unaudited: dependencies,
             policy: SortedMap::new(),
-            audit_as_crates_io: SortedMap::new(),
         }
     };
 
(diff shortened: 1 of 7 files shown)


Gankra:
This mostly-handles #193 but it doesn't implement the `regenerate` command or fix UX in general. That said, it should be good enough to unblock updating m-c as long as someone is willing to manually add the policy entries.

src/main.rs
```diff
@@ -1665,16 +1634,98 @@ fn foreign_packages<'a>(
...
     metadata: &'a Metadata,
     config: &'a ConfigFile,
 ) -> impl Iterator<Item = &'a Package> + 'a {
     // Only analyze things from crates.io (no source = path-dep / workspace-member)
     metadata
         .packages
         .iter()
-        .filter(|package| !package.is_third_party(&config.audit_as_crates_io))
+        .filter(|package| !package.is_third_party(&config.policy))
+}
```
mystor, line 1651: The comment suggests that this is supposed to ignore audit_as_crates_io, but it calls the `is_third_party` method, meaning it actually would take `audit_as_crates_io` into account, no?

Perhaps this method should pass an empty `&SortedMap::new()` in instead of the actual policy table?

It would be nice to have some tests for `check_audit_as_crates_io` to catch issues like this.

src/main.rs
```diff
@@ -1665,16 +1634,98 @@ fn foreign_packages<'a>(
...
+            .and_then(|policy| policy.audit_as_crates_io);
+        if audit_policy == Some(false) {
+            // They've explicitly said this is first-party so we don't care about what's in the registry
+            continue;
+        }
+
+        if let Some(index_entry) = cache.query_package_from_index(&package.name) {
+            for index_version in index_entry.versions() {
+                if let Ok(index_ver) = index_version.version().parse::<cargo_metadata::Version>() {
+                    if index_ver == package.version {
```
mystor, line 1676: FWIW I think that exposing a version of `query_package_from_index` which takes a name and a version might be nice (`query_package_version_from_index`?). When doing the checksum validation stuff, I had to add something like that as well, because I just wanted a specific version of the crate, and it's a bit fiddly to write out.

src/main.rs
```diff
@@ -1665,16 +1634,98 @@ fn foreign_packages<'a>(
...
+            out,
+            "error: some first-party packages match published crates.io versions"
+        )?;
+        for package in &needs_audit_as_entry {
+            writeln!(out, "  {}:{}", package.name, package.version)?;
+        }
+        writeln!(out)?;
+        writeln!(
+            out,
+            "All of these must have a policy.*.audit-as-crates-io entry"
```
mystor, line 1708: hyper-nit: I think it'd be nice to put the symbol `policy.*.audit-as-crates-io` in backticks or single quotes.

mystor (changes requested):
It would be really nice to have tests for `check_audit_as_crates_io` in isolation. It seems like the arguments are somewhat mockable, and the test machine in CI should have access to the index to build the crate in the first place.

A potential roadblock I see is that if we depend on a specific version of some crate not being posted theoretically someone could come by and post a version of that crate, but because we rely on a specific version, we can easily pick a crate (say `cargo-vet`) which we know will never release `version = "0.0.999-cargo.vet.tests.neverpublished"`.

We could also provide some way to mock out the crates-index crate for tests, but that's probably more effort.

src/main.rs
Gankra: I added this, but it has to be a two-pass thing since crates-index returns a wholy owned value for a Crate and subsequent Version iteration borrows from that value.

src/main.rs
Gankra: Good catch, fixed.

src/main.rs
mystor: You can clone the `Version` value and return it as an owned value. They shouldn't be that large or expensive to clone out and the UX improvement seems pretty nice.

Gankra:
Pushed up: added new testing infra for mocking the registry. A lot of it is stubbed out.

I also reached a breaking point and smashed the tests file into a dir with submodules for the different flavours of test. This sacrifices some of the spatial connections between different flavours of tests with the same basic input but I think that was barely useful.

src/main.rs
```diff
@@ -1254,7 +1254,9 @@ fn cmd_vet(out: &mut dyn Write, cfg: &Config) -> Result<(), VetError> {
         }
 
         // Check if any of our first-parties are in the crates.io registry
-        check_audit_as_crates_io(out, cfg, &store)?;
+        if check_audit_as_crates_io(out, cfg, &store).is_err() {
+            panic_any(ExitPanic(-1));
```
mystor, line 1258: Hmm, I don't love that this means we swallow any errors e.g. from `Cache::acquire` here. Could we make the test catch the panic and handle it that way instead?

mystor (approved):

Issue #175, opened 2022-06-13
investigate whether we should be using libgit2 over git-as-a-binary


Gankra:
We use git-as-a-binary for:

* get_user_info (global config user.email/user.name)
* diffing

This was done as just a quick-and-dirty thing while we were prototyping. It's possible we should be using libgit2 directly.

mystor:
I imagine we don't need the full functionality of libgit2, and can probably get away with just continuing to use the CLI, at least for now. The main thing which I could see being nice to move in-tree would be the diffing, but even if we pull in a library for that I think it'd be fine to use git to read basic configuration information.

Something like my (ab-?)use of GIT_EDITOR in #180 wouldn't be possible without the git CLI being available.

bholley:
I suppose I'll close this for now and we can reopen if we see a need.

mystor:
I do think this is more compelling as an option now that we depend on `git2` through `crates-index`, so we'd not be adding a new dependency.

PR #208, opened 2022-06-20
Fix: Auditor information is not read from global git config

matthiasbeyer:
The git configuration is used (in `get_user_info()` in src/main.rs) by
calling `git config --get`, which is not the same as reading the global
git configuration, as `git config` considers the repository-local
configuration file as well as the $HOME git configuration file.

Hence, adapt the wording here.

### DIFF
--- a/book/src/audit-entries.md
+++ b/book/src/audit-entries.md
@@ -53,7 +53,7 @@ dependencies](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.
 ## `who`
 
 A string identifying the auditor. When invoking `cargo vet certify`, the
-value is auto-populated from the global git config.
+value is auto-populated from the git config.
 
 This field is optional, but encouraged for two reasons:
 * It makes it easier to attribute audits at a glance, particularly for


matthiasbeyer:
On a seperate, but related note: Would you like to receive a PR for changing the git interfacing from calling `git` on CLI to usage of the `git2` crate?

mystor:
This was discussed a bit in #175. My current feelings are that that the benefits of using `git2` directly are somewhat unclear. We use git in the following ways right now:

1. We read the git configuration options to get the user's [name](https://github.com/mozilla/cargo-vet/blob/b94329f15a97551e1ac11f6f20c865ed4315c064/src/main.rs#L1530-L1538) and [email](https://github.com/mozilla/cargo-vet/blob/b94329f15a97551e1ac11f6f20c865ed4315c064/src/main.rs#L1530-L1538), which could potentially be simplified by using `git2`, though it's fairly simple already. 

2. We use it to get configuration information and invoke the user's [`GIT_EDITOR` to input notes](https://github.com/mozilla/cargo-vet/blob/d89cede93ddbf3d5c4e94cc2541fe822d71fb3cc/src/editor.rs#L48-L52). This is mostly useful as it's a somewhat reliable way to invoke an editor on Windows, and we'd get little benefit from using `git2` for this as we need the CLI around anyway, and we'd need to implement the logic behind `git var GIT_EDITOR` ourselves (as I don't think `git2` has the special handling).

3. We use it to diff two directories, neither of which is within a git repository. This is done to generate [diffstat information](https://github.com/mozilla/cargo-vet/blob/d89cede93ddbf3d5c4e94cc2541fe822d71fb3cc/src/storage.rs#L586-L596) for comparing and ranking deltas to suggest, and to [show to the user interactively](https://github.com/mozilla/cargo-vet/blob/b94329f15a97551e1ac11f6f20c865ed4315c064/src/main.rs#L1530-L1538) when they try to certify a delta. In the case of the full diff being shown to the user, we slightly benefit from using the git CLI, as it provides us an existing setup for starting a pager with the diff, as well as respecting the user's other diff config options, however it is somewhat inconvenient to work with in the diffstat case and a diffing API might be nice.

The biggest case which would be nice to streamline from those cases would be the diffstat case, and I don't think `git2` is the right tool for the job there. The `git2` crate is not threadsafe and does blocking I/O so it is awkward to integrate into our async tokio-based flow (you can see some of the struggle with that in my WIP patches for #206, which try to wrap the `cargo-index` crate (which uses `git2` internally)). In addition, AFAIK there's no actual mechanism in the API for diffing two directories unrelated to a repository, so it would require quite some hassle to get it working.

Because of that, I'm not sure that it's worth adding an explicit `git2` dependency just to read the author information from the git environment, and I'm inclined not to switch things to `git2` until we grow a more-significant use-case for it.

matthiasbeyer:
> The git2 crate is not threadsafe and does blocking I/O so it is awkward to integrate into our async tokio-based flow [...]

Right, that would be hard to get right.

I see that using git2 is not a good way to go here. Thanks for your very detailed answer! :+1:

PR #210, opened 2022-06-20
Improve criteria selection in certify

### LINKED ISSUE #192: Minimize reliance on command history for certify guessing
Now that we're making --guess-deeper the default in #190, there's no semantic ambiguity or performance risk in having `cargo vet certify` sometimes fall back to the suggest algorithm. As such we should just do that, and reduce the magical out-of-band reliance on previously-executed commands.

The only thing that should rely on that is bare `cargo vet certify`, which needs to guess which crate you're talking about.

mystor:
This patch makes 2 major changes to certify:

First, certify will no longer attempt to guess the version you are trying to certify unless there is a previous diff or inspect command in history for the crate being certified. This is to avoid potential mistakes, as it would be easy to miss the fact that the wrong version
was selected.

Second, when suggesting criteria to certify, the command will try harder to find the relevant set of criteria. If the diff or version
being certified isn't found in the `suggest` output history, it will now also check the unaudited table, and fall back to running the resolver to determine what criteria would heal the graph when specified for this revision.

This pass with the resolver is intended to be more relaxed that `suggest`, as the precise delta being checked is already known, so it
should be able to give suggested criteria even for audits which would not be suggested by the `suggest` algorithm.

Fixes #192

### DIFF
--- a/src/resolver.rs
+++ b/src/resolver.rs
@@ -2537,6 +2537,56 @@ impl<'a> ResolveReport<'a> {
         }))
     }
 
+    /// Given a package name and a delta to be certified, determine the set of
+    /// additional criteria for that delta/version pair which would have a
+    /// healing impact on the audit graph.
+    ///
+    /// This is more reliable than running a suggest and looking for a matching
+    /// output, as it will also select criteria for non-suggested audits.
+    pub fn compute_suggested_criteria(
+        &self,
+        package_name: PackageStr<'_>,
+        delta: &Delta,
+    ) -> Vec<CriteriaName> {
+        let fail = if let Conclusion::FailForVet(fail) = &self.conclusion {
+            fail
+        } else {
+            return Vec::new();
+        };
+
+        let mut criteria = self.criteria_mapper.no_criteria();
+
+        // Enumerate over the recorded failures, adding any criteria for this
+        // delta which would connect that package version into the audit graph.
+        for (&failure_idx, audit_failure) in &fail.failures {
+            let package = &self.graph.nodes[failure_idx];
+            if package.name != package_name {
+                continue;
+            }
+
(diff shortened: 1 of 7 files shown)


src/resolver.rs
```diff
@@ -2506,37 +2506,63 @@ impl<'a> ResolveReport<'a> {
...
+            for criteria_idx in audit_failure.criteria_failures.all().indices() {
+                let search_result = &result.search_results[criteria_idx];
+                if let SearchResult::Disconnected {
+                    reachable_from_root,
+                    reachable_from_target,
+                } = search_result
+                {
+                    if reachable_from_target.contains(&delta.to)
+                        && (delta.from == ROOT_VERSION || reachable_from_root.contains(&delta.from))
+                    {
```
Gankra, line 2553: Do we not add ROOT_VERSION to its own reachability set??? (The first part of the `||` seems pointless)

src/resolver.rs
```diff
@@ -2506,37 +2506,63 @@ impl<'a> ResolveReport<'a> {
...
             total_lines,
         }))
     }
 
+    /// Given a package name and a delta to be certified, determine the set of
+    /// additional criteria for that delta/version pair which would have a
+    /// healing impact on the audit graph.
+    ///
+    /// This is more reliable than running a suggest and looking for a matching
+    /// output, as it will also select criteria for non-suggested audits.
```
Gankra, line 2521: Interesting, I think I agree. Hooray for computing very precise results before even knowing what our actual goal is 🎉

Gankra (approved):
nice!

src/resolver.rs
mystor: I don't know, I just added in the explicit case to be extra sure. I can drop it and we can see if it still works I suppose

PR #212, opened 2022-06-21
Introduce an Out trait which abstracts over the program's output 

mystor:
This is certainly not a final solution. It doesn't integrate at all with our error reporting or tracing codepaths, and has the limitation that it requires mutable access thanks to using `io::Write` directly.

This allows us to mock out more of terminal interaction to allow it to be tested from within our unit tests.

This is then used when combined with a new ability to create mock cfgs from real argument lists to add a basic mocked-out test for the `certify` command.

This is currently built on top of #210, so contains the changes from that PR.


Gankra:
Gonna build some Error stuff on top of this

PR #218, opened 2022-06-22
Complete rework all of our error handling to favour explicit error ty…

Gankra:
…pes over printing and Report

This is a bit hacky because it's basically a first pass. After this we can refine the code more piecemeal.

### DIFF
--- a/src/editor.rs
+++ b/src/editor.rs
@@ -9,7 +9,7 @@ use std::str;
 use tempfile::NamedTempFile;
 use tracing::warn;
 
-use crate::VetError;
+use crate::errors::EditError;
 
 #[cfg(windows)]
 fn git_sh_path() -> Option<PathBuf> {
@@ -199,18 +199,18 @@ impl<'a> Editor<'a> {
 
     /// Run the editor, collecting and filtering the resulting file, and
     /// returning it as a string.
-    pub fn edit(self) -> Result<String, VetError> {
+    pub fn edit(self) -> Result<String, EditError> {
         // Close our handle on the file to allow other programs like the editor
         // to modify it on Windows.
         let path = self.tempfile.into_temp_path();
-        (self.run_editor)(&path)?;
+        (self.run_editor)(&path).map_err(EditError::CouldntLaunch)?;
 
         // Read in the result, filtering lines, and restoring unix line endings.
         // This is roughly based on git's logic for cleaning up commit message
         // files.
         let mut lines: Vec<String> = Vec::new();
-        for line in BufReader::new(File::open(&path)?).lines() {
-            let line = line?;
+        for line in BufReader::new(File::open(&path).map_err(EditError::CouldntOpen)?).lines() {
+            let line = line.map_err(EditError::CouldntRead)?;
             // Ignore lines starting with a comment character.
             if line.starts_with(self.comment_char) {
                 continue;
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -173,6 +173,12 @@ pub enum StoreAcquireError {
         FlockError,
     ),
     #[error(transparent)]
+    LoadToml(
+        #[from]
+        #[diagnostic(cause)]
+        LoadTomlError,
+    ),
+    #[error(transparent)]
     IoError(
         #[from]
         #[diagnostic(cause)]
@@ -202,6 +208,11 @@ pub enum StoreCommitError {
         #[from]
         #[diagnostic(cause)]
         std::io::Error,
+    ),
+    StoreToml(
+        #[from]
+        #[diagnostic(cause)]
+        StoreTomlError,
     ),
 }
 
@@ -259,6 +270,16 @@ pub enum CacheCommitError {
         #[diagnostic(cause)]
         std::io::Error,
     ),
+    StoreToml(
+        #[from]
+        #[diagnostic(cause)]
+        StoreTomlError,
+    ),
+    StoreJson(
+        #[from]
+        #[diagnostic(cause)]
+        StoreJsonError,
+    ),
 }
 
 //////////////////////////////////////////////////////////
@@ -303,12 +324,6 @@ pub enum DiffError {
         #[from]
         #[diagnostic(cause)]
         CommandError,
-    ),
-    #[error("failed to acquire permission to diff")]
-    DiffDenied(
-        #[from]
-        #[diagnostic(cause)]
-        tokio::sync::AcquireError,
     ),
 }
 
@@ -421,12 +436,6 @@ pub enum FetchAuditError {
 #[derive(Debug, Error, Diagnostic)]
 #[non_exhaustive]
 pub enum DownloadError {
-    #[error("failed to acquire permission to download {url}")]
-    DonwloadDenied {
-        url: reqwest::Url,
-        #[diagnostic(cause)]
-        error: tokio::sync::AcquireError,
-    },
     #[error("failed to start download of {url}")]
     FailedToStartDownload {
         url: reqwest::Url,
@@ -493,7 +502,7 @@ pub enum FlockError {
 }
 
 //////////////////////////////////////////////////////////
-// TomlParseError
+// TomlError/JsonError
 //////////////////////////////////////////////////////////
 
 #[derive(Debug, Error, Diagnostic)]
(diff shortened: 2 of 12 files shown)


Gankra:
image

Hmm hallucinating related...

Gankra:
image

src/editor.rs
```diff
@@ -9,7 +9,7 @@ use std::str;
 use tempfile::NamedTempFile;
 use tracing::warn;
 
-use crate::VetError;
+use crate::errors::EditError;
```
mystor, line 12: We might want to eventually change the `io::ERror` returning methods to return `EditError`, but this is fine for now.

src/errors.rs
```diff
@@ -0,0 +1,508 @@
...
+        error: std::io::Error,
+    },
+    #[error("Failed to finalize unpack to {}", target.display())]
+    LockCreate {
+        target: std::path::PathBuf,
+        #[diagnostic(cause)]
+        error: std::io::Error,
+    },
+    #[error(transparent)]
+    IoError(#[from] std::io::Error),
```
mystor, line 388: I'm not sure that we need to have this much granularity for errors when unpacking crates, we can probably get away with coalescing them somewhat as it's not like the user can do anything about a package downloaded from crates.io being broken.

src/errors.rs
```diff
@@ -0,0 +1,508 @@
...
+
+#[derive(Debug, Error, Diagnostic)]
+#[non_exhaustive]
+pub enum DownloadError {
+    #[error("failed to acquire permission to download {url}")]
+    DonwloadDenied {
+        url: reqwest::Url,
+        #[diagnostic(cause)]
+        error: tokio::sync::AcquireError,
+    },
```
mystor, line 429: This should be impossible as we never close the semaphore. Perhaps we should just unwrap that error?

I worry that this message is misleading as-is right now if it did fire, because "permission to download" sounds like perhaps a HTTP 403 Forbidden or filesystem permission error, which it isn't.

src/flock.rs
```diff
@@ -130,9 +129,10 @@ impl Filesystem {
     ///
     /// Handles errors where other processes are also attempting to concurrently
     /// create this directory.
-    pub fn create_dir(&self) -> Result<(), VetError> {
-        fs::create_dir_all(&self.root)
-            .wrap_err_with(|| format!("failed to create directory `{}`", self.display()))
+    pub fn create_dir(&self) -> Result<(), std::io::Error> {
+        fs::create_dir_all(&self.root)?;
+        Ok(())
+        //    .wrap_err_with(|| format!("failed to create directory `{}`", self.display()))
```
mystor, line 135: This context wrapping was actually inherited from cargo because the errors in the stdlib don't include the directory path information. Probably not a huge deal to drop though.

Let's drop the `?;\nOk(())` and the commented out code though?

src/flock.rs
```diff
@@ -256,7 +251,7 @@ fn acquire(
     path: &Path,
     lock_try: &dyn Fn() -> io::Result<()>,
     lock_block: &dyn Fn() -> io::Result<()>,
-) -> Result<(), VetError> {
+) -> Result<(), std::io::Error> {
```
mystor, line 254: It feels like this ought to return a `FlockError` if there's any code which returns one here, though perhaps we can get away with just using `io::Error` everywhere in `flock`.

src/flock.rs
```diff
@@ -281,13 +276,15 @@ fn acquire(
 
         Err(e) => {
             if !error_contended(&e) {
-                return Err(e).wrap_err(format!("failed to lock file: {}", path.display()));
+                // .wrap_err(format!("failed to lock file: {}", path.display()))
```
mystor, line 279: Let's not leave around commented out `wrap_err` calls.

src/flock.rs
```diff
@@ -281,13 +276,15 @@ fn acquire(
...
+                Err(e)?;
             }
         }
     }
     eprintln!("Blocking: waiting for file lock on {}", msg);
 
-    lock_block().wrap_err_with(|| format!("failed to lock file: {}", path.display()))?;
+    // format!("failed to lock file: {}", path.display())
```
mystor, line 286: It'd be really nice to keep having errors like this from flock, could we perhaps consider making this method return `FlockError` and adding some special cases for this?

src/main.rs
```diff
@@ -107,7 +112,7 @@ struct ExitPanic(i32);
 /// as our final act.
 struct ExecPanic(std::process::Command);
 
-fn main() -> Result<(), VetError> {
+fn main() -> Result<(), ()> {
```
mystor, line 115: I don't understand why we're bypassing the normal termination handling here.

That being said, if we want to handle every result and `std::process::exit` manually, we can have `main()` return nothing, and handle the error ourself I suppose.

It feels to me like we should just return a `miette::Report` here.

src/main.rs
```diff
@@ -1180,23 +1195,26 @@ fn cmd_diff(out: &mut dyn Out, cfg: &PartialConfig, sub_args: &DiffArgs) -> Resu
...
+        .block_on(async {
+            tokio::try_join!(
+                cache.fetch_package(network.as_ref(), package, &sub_args.version1),
+                cache.fetch_package(network.as_ref(), package, &sub_args.version2)
+            )
+        })
+        .into_diagnostic()?;
 
-    writeln!(out)?;
+    writeln!(out).into_diagnostic()?;
```
mystor, line 1210: Is this `into_diagnostic()` really required?

src/main.rs
```diff
@@ -1314,18 +1333,19 @@ fn cmd_help_md(
...
+    writeln!(out, "# {pretty_app_name} CLI manual").into_diagnostic()?;
+    writeln!(out).into_diagnostic()?;
     writeln!(
         out,
         "> This manual can be regenerated with `{pretty_app_name} help-markdown`"
-    )?;
-    writeln!(out)?;
+    )
+    .into_diagnostic()?;
+    writeln!(out).into_diagnostic()?;
```
mystor, line 1348: This is super duper gross.

src/network.rs
```diff
@@ -54,25 +53,57 @@ impl Network {
...
+            OsStr::new(".part"),
+        ]));
         {
-            let _permit = self.connection_semaphore.acquire().await?;
+            let _permit = self.connection_semaphore.acquire().await.map_err(|error| {
+                DownloadError::DonwloadDenied {
+                    url: url.clone(),
+                    error,
+                }
+            })?;
```
mystor, line 71: This error really should never happen. We should just `.unwrap()` it.

src/network.rs
```diff
@@ -81,32 +112,45 @@ impl Network {
...
     /// Download a file into memory
-    pub async fn download(&self, url: Url) -> Result<Vec<u8>, VetError> {
-        let _permit = self.connection_semaphore.acquire().await?;
+    pub async fn download(&self, url: Url) -> Result<Vec<u8>, DownloadError> {
+        let _permit = self.connection_semaphore.acquire().await.map_err(|error| {
+            DownloadError::DonwloadDenied {
+                url: url.clone(),
+                error,
+            }
+        })?;
```
mystor, line 132: We should unwrap this.

src/network.rs
```diff
@@ -54,25 +53,57 @@ impl Network {
     }
 
     /// Download a file and persist it to disk
-    pub async fn download_and_persist(&self, url: Url, persist_to: &Path) -> Result<(), VetError> {
-        let download_tmp_path = OsString::from_iter([persist_to.as_os_str(), OsStr::new(".part")]);
+    pub async fn download_and_persist(
```
mystor, line 56: I feel like with some functions like this what we really want is to apply some wrapper to the error as it leaves this function, but we really don't need this much granularity for each piece of error in the function.

src/storage.rs
```diff
@@ -549,20 +551,25 @@ impl Cache {
                         file
                     })
                 })
-                .await?;
+                .await
+                .expect("failed to join");
```
mystor, line 555: This join error is what happens if the spawned task panics, like when joining a thread.

src/storage.rs
```diff
@@ -1009,46 +1027,46 @@ fn find_cargo_registry() -> Result<CargoRegistry, VetError> {
...
     Ok(json)
 }
-fn store_json<T>(mut writer: impl Write, val: T) -> Result<(), VetError>
+fn store_json<T>(mut writer: impl Write, val: T) -> Result<(), std::io::Error>
 where
     T: Serialize,
 {
     // FIXME: do this in a temp file and swap it into place to avoid corruption?
-    let json_string = serde_json::to_string(&val)?;
+    let json_string = serde_json::to_string(&val).unwrap();
```
mystor, line 1064: This also seems like a behaviour regression from before.

mystor (commented):
My general thoughts here are that I think we're being perhaps a bit too granular with our errors in some places, where we should perhaps instead have a single toplevel error for the method, and hide the precise ways in which it can fail internally a bit more. If we were to make this a library, for example, we wouldn't want to expose in our error types that we shell out to a subcommand to implement `diffstat`, as that's an implementation detail. We can just have a `Box<dyn Error>` or similar in those situations as a cause, with some metadata about why we're failing.

I think technically speaking these may also harm happy-path performance in some cases, as we're expanding the error side of the enum quite a bit, but I don't see a clear easy alternative, so I think we can go with it.

I pointed out a few places where I worry we're making the recovery worse (specifically around parsing, which I think we should fix before merging), and I really don't like the conversion from `writeln!` calls (but we can fix that by changing `Out` to have an infallible `write_fmt` method rather than just inheriting from `io::Write`, as we don't really need to worry about those errors anyway).

src/tests/snapshots/cargo_vet__tests__audit_as_crates_io__simple-audit-as-crates-io.snap
```diff
@@ -2,8 +2,11 @@
...
-error: some non-crates.io-fetched packages match published crates.io versions
-  first-party:10.0.0
 
-All of these must have a `policy.*.audit-as-crates-io` entry
+  × There are some issues with your policy.audit-as-crates-io entries
+
+Error: 
+  × Some first-party packages match published crates.io versions
+  │   first-party:10.0.0
+  help: Add a `policy.*.audit-as-crates-io` entry for them
```
str4d, line 11: The change in this PR reverted https://github.com/mozilla/cargo-vet/commit/a00b95ae863eb214f7d907c05ae0475a4079e876, possibly unintentionally?

bholley: Good catch, I just pushed https://github.com/mozilla/cargo-vet/commit/aaaf4f7b964c89ac7e6e7551b37a6bbe71be1e00 to switch it back.

PR #221, opened 2022-06-23
Make read_line_initial_text omit prompt from the returned input.

jimblandy:
Don't pass the prompt string to `console::read_line_initial_text`, as that causes it to include the prompt in the input: entering "2" and hitting ENTER returns the string "> 2". Just write the prompt to the console, and then call read_line_initial_text with a null string.

### DIFF
--- a/src/out.rs
+++ b/src/out.rs
@@ -25,7 +25,7 @@ pub trait Out: io::Write {
 
     /// Ask the user a question, and read in a line with the user's response. If
     /// there's no user able to respond, `None` will be returned instead.
-    fn read_line_initial_text(&mut self, _initial: &str) -> io::Result<String> {
+    fn read_line_with_prompt(&mut self, _prompt: &str) -> io::Result<String> {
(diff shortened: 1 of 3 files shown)


src/out.rs
```diff
@@ -53,8 +53,10 @@ impl Out for Term {
         (&*self).clear_screen()
     }
 
-    fn read_line_initial_text(&mut self, initial: &str) -> io::Result<String> {
-        (&*self).read_line_initial_text(initial)
+    fn read_line_with_prompt(&mut self, prompt: &str) -> io::Result<String> {
+        self.write_str(prompt)?;
+        self.flush()?;
+        (&*self).read_line_initial_text("")
```
mystor, line 59: You can just call `read_line()` instead: https://docs.rs/console/latest/console/struct.Term.html#method.read_line

mystor:
(thanks for the catch! @Gankra will merge this while I'm out)

Gankra (approved):
r=me with nika's requested change

PR #234, opened 2022-06-27
Interactively display criteria and instructions for inspect and diff

### LINKED ISSUE #200: Interactively display criteria and instructions when invoking inspect/diff
When someone encounters a cargo-vet failure for the first time, the output will tell them to run `inspect` or `diff` and indicate which criteria to audit for, but there's never an opportunity for the person to read the actual criteria until they're done with the audit and trying to certify. This is obviously not ideal for first-timers.

My proposal here is that `inspect` and `diff` should infer the criteria (with the same algorithm that `certify` uses), and print the description(s) to the terminal, perhaps along with some additional helpful instructions ("these audits impact the ecosystem", "ask for help if you're not sure", etc). The user then hits `enter` to continue with the inspect/diff (which currently results in more activities in the terminal, but will soon open up a web inspection view unless --local is passed).

We can then add a --non-interactive flag to skip the above.

mystor:
This changes the behaviour of the inspect and diff instructions to now
predict the required criteria for the audit, and display the relevant
information up-front. A warning is shown if no criteria could be
determined, to help avoid accidentally unnecessary audits.

In addition, a partial audit for `syn` was added to the test project to
make it easier to test delta audit suggestions.

Fixes #200

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -1,18 +1,15 @@
 use std::collections::HashMap;
-use std::mem;
 use std::ops::Deref;
 use std::panic::panic_any;
-use std::path::Path;
 use std::time::Duration;
-use std::{fs::File, panic, path::PathBuf};
+use std::{fs::File, io, mem, panic, path::PathBuf};
 
-use cargo_metadata::{Metadata, Package};
+use cargo_metadata::{Metadata, Package, Version};
 use clap::{CommandFactory, Parser};
 use console::Term;
 use errors::{
-    AuditAsError, AuditAsErrors, CertifyError, DiffError, MinimizeUnauditedError,
-    NeedsAuditAsError, NeedsAuditAsErrors, ShouldntBeAuditAsError, ShouldntBeAuditAsErrors,
-    UserInfoError,
+    AuditAsError, AuditAsErrors, CertifyError, MinimizeUnauditedError, NeedsAuditAsError,
+    NeedsAuditAsErrors, ShouldntBeAuditAsError, ShouldntBeAuditAsErrors, UserInfoError,
 };
 use format::{CriteriaName, CriteriaStr, PackageName, PolicyEntry};
 use futures_util::future::join_all;
@@ -20,7 +17,7 @@ use miette::{miette, Context, IntoDiagnostic};
 use network::Network;
 use reqwest::Url;
 use serde::de::Deserialize;
-use tracing::{error, info, trace};
+use tracing::{error, info, trace, warn};
 
 use crate::cli::*;
 use crate::errors::{CommandError, DownloadError};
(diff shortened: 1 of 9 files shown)


src/main.rs
```diff
@@ -422,8 +417,9 @@ fn real_main() -> Result<(), miette::Report> {
         Some(FetchImports(sub_args)) => cmd_fetch_imports(out, &cfg, sub_args),
         Some(RegenerateUnaudited(sub_args)) => cmd_regenerate_unaudited(out, &cfg, sub_args),
         Some(DumpGraph(sub_args)) => cmd_dump_graph(out, &cfg, sub_args),
-        // Need to be non-exhaustive because freestanding commands were handled earlier
-        _ => unreachable!("did you add a new command and forget to implement it?"),
+        Some(Inspect(sub_args)) => cmd_inspect(out, &cfg, sub_args),
+        Some(Diff(sub_args)) => cmd_diff(out, &cfg, sub_args),
```
Gankra, line 421: Are we ok with regressing #29? (not regressing it is a frustrating problem to solve...)

src/out.rs
```diff
@@ -24,11 +24,22 @@ pub trait Out: io::Write {
...
+    /// Like `read_line_with_prompt`, except will be run asynchronously when
+    /// possible in order to avoid blocking other futures running on the current
+    /// thread. This can be used to allow running background tasks, such as
+    /// fetching resources, while waiting for the user to respond.
+    fn read_line_with_prompt_async<'a>(
+        &'a mut self,
+        initial: &'a str,
+    ) -> Pin<Box<dyn Future<Output = io::Result<String>> + 'a>> {
+        Box::pin(async { self.read_line_with_prompt(initial) })
+    }
```
Gankra, line 41: neat

Gankra (approved):

src/out.rs
mystor: This doesn't actually make it async, it's 100% a cheat for cases where we can't make it async

src/main.rs
bholley: I would prefer not to regress it. How much work is it to handle?

Gankra: Currently we have a hard distinction between "commands which can access the store and therefore require one" and "commands which can't access the store" with a type-level distinction on the kind of "Config" that gets passed in. This would require adding the ability to detect whether the store exists ahead of time and optionally enabling features based on that.

Gankra: (Which would completely change the way main/config is handled)

bholley: I think I'm actually OK with regressing it for now, which will also ensure people don't unintentionally miss our guidance flow.

PR #235, opened 2022-06-28
Allow specifying global options after subcommands

### LINKED ISSUE #94: figure out global-arg-cascading-to-subcommands
right now `cargo vet suggest --output-file=output.txt` doesn't work and you have to write `cargo vet --output-file=output.txt suggest`. This is an artifact of how I hang the global args off of the main command. Part of me kind of likes that it works this way but I know it's bad and should be fixed.

Figure out what the "best" way is to have the global args available on every subcommand.

Figure out whether docs for each subcommand should list off all the global args, or if we should try to suppress them.

mystor:
This patch makes a number of changes to subcommand parsing and help formatting to allow global options to be specified on subcommands as well as on the toplevel command. In addition, the `cargo vet` default operation is now also mapped to a `cargo vet check` subcommand.

The cargo workspace command line options were removed, as the `cargo metadata` command never accepted them, so previously passing them would lead to `cargo vet` exiting with an error.

This is a breaking change to the CLI interface, as global options must now be specified after the subcommand rather than before it.

The new markdown help rendering can be previewed here: https://gist.github.com/mystor/ab1ecbbfa01bd873dee1f021ce50340d

Fixes #94


Gankra (approved):
Seems solid?

PR #237, opened 2022-06-29
Cli blitz

### LINKED ISSUE #193: Move audit-as-crates-io to the policy table and implement `cargo vet regenerate`

### LINKED ISSUE #228: Rename the unaudited table to the "exemptions" table
I've concluded that the current name is confusing, and that it's more self-explanatory to talk about "audits, imports, and exemptions".

Given that we haven't done a proper release yet and things are still somewhat use-at-your-own-risk, I don't think an automatic migration is essential. But @Gankra thinks it might be easy to do with a serde annotation — if so we might as well do that.

Gankra:
Fixes #228 
Fixes #193 

* first commit: mostly doc/cli changes that get cleaned up in subsequent commits. (initial shifts to "exemptions" happens here and is smeared through the commits, but most changes are in the last commit)

* second commit: `regenerate audit-as-crates-io` impl, just runs check_audit_as_crates_io and sets all erroring policies to `Some(false)`. This is now implicitly performed by `init` so that `check` will be guaranteed to pass again.

* third commit: `regenerate imports` impl. This is identical to `fetch imports` except it sets an `accept_changes` flag to true. The bulk of the change here is to execute `eula_for_criteria` while fetching the imports to cache the value in `description` if it has a `description_url` and to check if it differs from the previously cached value (if it exists). The `accept_changes` flag determines if a mismatch is an error.

* fourth commit: "exemptions" rename. I avoided renaming some of the test files to minimize the diff on the tests. We can finish that in a second pass.

The second and third commit both merit me writing some tests for that functionality. The second is kinda trivial, while the third would require me setting up a new kind of test that can mock the fetch..?

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -355,15 +355,28 @@ pub enum Commands {
 
 #[derive(Subcommand)]
 pub enum RegenerateSubcommands {
+    /// Regenerate your exemptions to make `check` pass minimally
     ///
+    /// This command can be used for two purposes: to force your supply-chain to pass `check`
+    /// when it's currently failing, or to minimize/garbage-collect your exemptions when it's
+    /// already passing. These are ultimately the same operation.
+    ///
+    /// We will try our best to preserve existing exemptions, removing only those that
+    /// aren't needed, and adding only those that are needed. Exemptions that are overbroad
+    /// may also be weakened (i.e. safe-to-deploy may be reduced to safe-to-run).
     #[clap(disable_version_flag = true)]
     Exemptions(RegenerateExemptionsArgs),
 
-    /// Suggest some low-hanging fruit to review
+    /// Regenerate your imports and accept changes to criteria
+    ///
+    /// This is equivalent to `cargo vet fetch-imports` but it won't produce an error if
+    /// the descriptions of foreign criteria change.
     #[clap(disable_version_flag = true)]
     Imports(RegenerateImportsArgs),
 
-    /// Initialize cargo-vet for your project
+    /// Regenerate you audit-as-crates-io entries to make `check` pass
+    ///
+    /// This will just set any problematic entries to `audit-as-crates-io = false`.
     #[clap(disable_version_flag = true)]
     AuditAsCratesIo(RegenerateAuditAsCratesIoArgs),
 }
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -198,7 +198,7 @@ pub enum Commands {
     /// If you don't consider an exemption to be "backlog", add `suggest = false` to its
     /// entry and we won't remove it while suggesting.
     ///
-    /// See also `regenerate unaudited`, which can be used to "garbage collect"
+    /// See also `regenerate exemptions`, which can be used to "garbage collect"
     /// your backlog (if you run it while `check` is passing).
     #[clap(disable_version_flag = true)]
     Suggest(SuggestArgs),
@@ -474,10 +474,10 @@ pub struct RecordViolationArgs {
 /// Certifies the given version
 #[derive(clap::Args)]
 pub struct AddExemptionArgs {
-    /// The package to mark as unaudited (trusted)
+    /// The package to mark as exempted
     #[clap(action)]
     pub package: PackageName,
-    /// The version to mark as unaudited
+    /// The version to mark as exempted
     #[clap(action)]
     pub version: Version,
     /// The criteria to assume (trust)
@@ -485,7 +485,7 @@ pub struct AddExemptionArgs {
     /// If not provided, we will prompt you for this information.
     #[clap(long, action)]
     pub criteria: Vec<CriteriaName>,
-    /// The dependency-criteria to require for this unaudited entry to be valid
+    /// The dependency-criteria to require for this exemption to be valid
     ///
     /// If not provided, we will still implicitly require dependencies to satisfy `criteria`.
     #[clap(long, action)]
@@ -495,7 +495,7 @@ pub struct AddExemptionArgs {
     /// If not provided, there will be no notes.
     #[clap(long, action)]
     pub notes: Option<String>,
-    /// Suppress suggesting this unaudited entry
+    /// Suppress suggesting this exemption for review
     #[clap(long, action)]
     pub no_suggest: bool,
 }
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -154,77 +154,218 @@ pub struct Cli {
 
 #[derive(Subcommand)]
 pub enum Commands {
+    // Main commands:
     /// \[default\] Check that the current project has been vetted
     ///
     /// This is the default behaviour if no subcommand is specified.
+    ///
+    /// If the check fails due to lack of audits, we will do our best to explain why
+    /// vetting failed, and what should be done to fix it. This can involve a certain
+    /// amount of guesswork, as there are many possible solutions and we only want to recommend
+    /// the "best" one to keep things simple.
+    ///
+    /// Failures and suggestions can either be "Certain" or "Speculative". Speculative items
+    /// are greyed out and sorted lower to indicate that the Certain entries should be looked
+    /// at first. Speculative items are for packages that probably need audits too, but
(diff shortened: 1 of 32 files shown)


tests/snapshots/test_cli__markdown-help.snap
```diff
@@ -304,12 +371,110 @@ Print help information
...
+#### `-h, --help`
+Print help information
+
+### GLOBAL OPTIONS
+This subcommand accepts all the [global options](#global-options)
+### SUBCOMMANDS
+* [exemptions](#cargo-vet-exemptions): 
+* [imports](#cargo-vet-imports): Suggest some low-hanging fruit to review
+* [audit-as-crates-io](#cargo-vet-audit-as-crates-io): Initialize cargo-vet for your project
+* [help](#cargo-vet-help): Print this message or the help of the given subcommand(s)
```
Gankra, line 398: This is a usable result, but we should probably be namespacing the anchors properly.

tests/snapshots/test_cli__markdown-help.snap
```diff
@@ -304,20 +371,118 @@ Print help information
...
```
Gankra, line 398: this is a usable result, but we should probably properly namespace the anchors...

src/cli.rs
```diff
@@ -154,79 +154,220 @@ pub struct Cli {
...
-    /// Mark `$package $version` as a violation of policy
+    /// Mark a package as exempted from review
+    ///
+    /// Exemptions are *usually* just "backlog" and the expectation is that you will review
+    /// them "eventually". You should usually only be trying to remove them, but sometimes
+    /// additions are necessary to make progress.
+    ///
+    /// `regenerate exemptions` will do this for your automatically to make `check` pass
+    /// (and remove any unnecessary ones), so we recommend using that over `add-exemption`.
+    /// This command mostly exists as "plumbing" for building tools on top of `cargo vet`.
```
mystor, line 273: If this is a plumbing command, perhaps it should be in the plumbing commands section below?

src/cli.rs
```diff
@@ -154,79 +154,220 @@ pub struct Cli {
...
     /// any extra files which aren't recognized by cargo-vet.
     ///
     /// In the future, many cargo-vet subcommands will implicitly do this.
     #[clap(disable_version_flag = true)]
     Gc(GcArgs),
 }
 
+#[derive(Subcommand)]
+pub enum RegenerateSubcommands {
+    ///
```
mystor, line 358: This should probably have a non-empty doc comment

```suggestion
    /// Regenerate the `exemptions` list, removing unnecessary crates, and
    /// adding new ones such that `cargo vet check` will pass.
```

src/cli.rs
```diff
@@ -154,79 +154,220 @@ pub struct Cli {
...
+    #[clap(disable_version_flag = true)]
+    Exemptions(RegenerateExemptionsArgs),
+
+    /// Suggest some low-hanging fruit to review
+    #[clap(disable_version_flag = true)]
+    Imports(RegenerateImportsArgs),
+
+    /// Initialize cargo-vet for your project
```
mystor, line 366: This isn't an accurate comment.

src/main.rs
```diff
@@ -392,22 +393,24 @@ fn real_main() -> Result<(), miette::Report> {
...
         Some(Fmt(sub_args)) => cmd_fmt(out, &cfg, sub_args),
         Some(FetchImports(sub_args)) => cmd_fetch_imports(out, &cfg, sub_args),
-        Some(RegenerateUnaudited(sub_args)) => cmd_regenerate_unaudited(out, &cfg, sub_args),
         Some(DumpGraph(sub_args)) => cmd_dump_graph(out, &cfg, sub_args),
         Some(Inspect(sub_args)) => cmd_inspect(out, &cfg, sub_args),
         Some(Diff(sub_args)) => cmd_diff(out, &cfg, sub_args),
         Some(HelpMarkdown(_)) | Some(Gc(_)) => unreachable!("handled earlier"),
+        Some(Regenerate(Imports(sub_args))) => cmd_regenerate_imports(out, &cfg, sub_args),
+        Some(Regenerate(Exemptions(sub_args))) => cmd_regenerate_exemptions(out, &cfg, sub_args),
+        Some(Regenerate(AuditAsCratesIo(sub_args))) => cmd_regenerate_audit_as(out, &cfg, sub_args),
```
mystor, line 413: Let's put these above the `unreachable!(` arm.

src/main.rs
```diff
@@ -1140,31 +1145,118 @@ fn cmd_suggest(
...
+    }
+
+    // ERRORS: just a warning that you're holding it wrong, unclear if immediate or buffered,
+    // or if this should be a hard error, or if we should ignore the --locked flag and
+    // just do it anyway
+    writeln!(
+        out,
+        "warning: ran `regenerate imports` with --locked, this won't do anything!"
+    )
+    .into_diagnostic()?;
```
mystor, line 1176: Should this perhaps be a `warn!` through tracing? I suppose `fetch-imports` does it this way as well.

src/storage.rs
```diff
@@ -427,18 +427,72 @@ impl Store {
...
+        network: &Network,
+        accept_changes: bool,
+    ) -> Result<(), FetchAuditError> {
+        let raw_new_imports =
+            try_join_all(self.config.imports.iter().map(|(name, import)| async {
                 let audit_file = fetch_foreign_audit(network, name, &import.url).await?;
-                Ok::<_, FetchAuditError>((name.clone(), audit_file))
+                // Fetch the descriptions to cache them and check that they haven't changed
+                // FIXME: this should probably treat failing to fetch as an error but eula_for_criteria
+                // hides errors... should we have two versions? Or make it the caller's problem?
```
mystor, line 440: Probably should make it the caller's problem, and then perhaps add a special infallible version?

src/tests/regenerate_unaudited.rs
```diff
@@ -1,36 +1,36 @@
 use super::*;
```
mystor, line 1: This file should probably be renamed to `regenerate_exemptions.rs`

mystor (approved):
Please fix the doc comments for the regenerate subcommands, but otherwise this looks good :-)

We can rename the test module in a follow-up because it'll definitely cause a lot of churn.

tests/snapshots/test_cli__markdown-help.snap
mystor: Yeah, that would probably be a good idea. Perhaps change the todo list to include a `(String, &mut clap::Command)` instead, so that you can pass the "full name" of the command alongside it?

src/cli.rs
Gankra: It exists in a twilight between plumbing and procelain, so it sits on the boundary between the two (which otherwise have no user-facing labels, only ordering-by-convention)

src/main.rs
Gankra: Yeah I'm just reproducing the existing issue.

src/tests/regenerate_unaudited.rs
Gankra: yeah wanted to defer that

src/storage.rs
Gankra: Gonna leave this to a followup

Issue #245, opened 2022-06-30
Add a --mode to inspect and diff and implement sourcegraph support


bholley:
Initially supported will be `local` (the current behavior) and `sourcegraph`

For sourcegraph, we should open a browser tab (after acknowledging the interactive part). Inspect should route like [so](https://sourcegraph.com/crates/generic-array@v0.14.5) and diff should route like [so](https://sourcegraph.com/crates/generic-array/-/compare/v0.14.4...v0.14.5).

mystor:
Fixed by #246 and #247

PR #246, opened 2022-06-30
Add --mode to inspect and diff, add support for sourcegraph urls

Gankra:


Gankra:
fixes #245

src/main.rs
```diff
@@ -543,6 +543,28 @@ fn cmd_inspect(
...
+                None,
+                version,
+            ))
+            .into_diagnostic()?;
+
+        let url = format!("https://sourcegraph.com/crates/{package}@v{version}");
+        match open::that(&url) {
+            Ok(()) => return Ok(()),
+            Err(e) => {
+                warn!("Couldn't open browser, falling back to local {:?}", e);
```
mystor, line 563: This fallback behaviour is kinda terrible, given that if we hit it we'll prompt the user to accept EULAs again... It would be nice if we could hold ourselves to only trying to do that once. Perhaps we could define a local closure like:

```rust
let mut did_prompt_eulas = false;
let mut prompt_eulas = || async {
  if did_prompt_eulas {
    return Ok(());
  }
  did_prompt_eulas = true;
  return prompt_criteria_eulas(out, cfg, network.as_ref(), &store, package, None, version).await;
};
```

src/main.rs
```diff
@@ -543,6 +543,28 @@ fn cmd_inspect(
...
+                &store,
+                package,
+                None,
+                version,
+            ))
+            .into_diagnostic()?;
+
+        let url = format!("https://sourcegraph.com/crates/{package}@v{version}");
+        match open::that(&url) {
+            Ok(()) => return Ok(()),
```
mystor, line 561: Let's log something to the console to let the user know we opened this page in their browser. In some systems, the user might be running such that the browser isn't brought to the front when they open it, and it could be quite surprising when the UI just doesn't show up because e.g. firefox stayed in the background, and the command just immediately exited.

I know I personally would be running in a situation where it might be weird, as I run commands on a remote SSH server, so any browser started will not be visible to me at all (as it's on a different computer)

mystor (changes requested):

src/main.rs
```diff
@@ -1414,6 +1456,36 @@ fn cmd_diff(out: &Arc<dyn Out>, cfg: &Config, sub_args: &DiffArgs) -> Result<(),
...
+        match open::that(&url) {
+            Ok(()) => {
+                writeln!(out, "opened {url} in your browser");
+                return Ok(());
+            }
+            Err(_e) => {
+                return Err(miette!(
+                    "Couldn't open {url} in your browser, try --mode=local?"
+                ));
+            }
```
mystor, line 1485: I think we want to acknowledge the original cause here, because it might be useful. Can we `wrap_err` instead of throwing out the old error?

src/main.rs
```diff
@@ -1414,6 +1456,36 @@ fn cmd_diff(out: &Arc<dyn Out>, cfg: &Config, sub_args: &DiffArgs) -> Result<(),
...
+                package,
+                Some(version1),
+                version2,
+                prompt,
+            ))
+            .into_diagnostic()?;
+
+        match open::that(&url) {
+            Ok(()) => {
+                writeln!(out, "opened {url} in your browser");
```
mystor, line 1478: I think this is probably unnecessary given that we just prompted the user to agree to open the URL in their browser, and they hit ENTER, but it's fine.

PR #257, opened 2022-07-11
remove references to unimplemented 'targets' handling

### LINKED ISSUE #249: remove "target" stuff for now
it's unimplemented and we don't have a good design for it. we should just remove all references to it to avoid confusion.

mystor:
This wasn't implemented, and it's not entirely clear how it would behave, so we're removing references to it to avoid confusion.

Fixes #249


Gankra:
rad thx

PR #261, opened 2022-07-13
Store verbatim source in imports.lock

mystor:
This patch changes how imports.lock is handled internally, such that verbatim imported sources are now generally preserved for error reporting purposes, and information like criteria descriptions are consistently cached locally.

With this change, the imports.lock is now a list of URIs which have been fetched and imported into this repository's supply-chain. This is used for both imported criteria descriptions and imported audits. When loading the store, imports are now resolved, being injected into the relevant places where they were mentioned in the `ConfigFile` and `AuditsFile` data structures so that other code can use then easily.

The handling of criteria descriptions in this way was done to avoid needing to modify the imported source text, so that errors can continue to be reported relative to the original imported source text.

Each imported URL starts with a block of TOML "front-matter" (delimited by `---` lines) which describes the URL and the number of lines. This was used rather than a more concise format to allow recording cache keys, metadata, or other sources in the future. The ['hrx' format](https://github.com/google/hrx) was considered for this file, however was not used, as the design is meant to model archives and filesystem paths, meaning that URIs are not legal file names (due to containing a ':' character).

After this change, the first run of `cargo vet` on your repository will report a warning, and your `imports.lock` file will need to be re-generated. This should happen automatically.

This PR also includes the following other changes to make this possible:

* **Use a more transparent type for source files**

   This patch introduces our own implementation of miette's `SourceCode` trait which allows accessing the string value and holds the name. This is then used instead of `NamedSource` to allow things like parsers to run on an `Arc<SourceFile>`, reducing string copies.

   In addition, this `SourceFile` kind has support for doing nested files, which will be used to avoid unnecessary string copies when reading files from the `imports.lock` file.

 * **Also set implied criteria in CriteriaMapper::set_criteria**

   Without this change a mapped criteria of e.g. "safe-to-deploy" ->"safe-to-deploy" would not actually fulfil any requirements due to "safe-to-run" not being implied by the mapped criteria.


Gankra (approved):
great stuff, sorry it took me so long

mystor:
Based on discussions in #272 and on matrix, we probably don't want to keep the full verbatim source in imports.lock, so I'm going to close this out for now. We might take parts of this and land it seperately.

PR #273, opened 2022-07-20
namespace and map criteria properly

### LINKED ISSUE #240: reimplement foreign audit handling in resolver
this code was hacked up in like week 1 and hasn't been touched since.

The main challenge to solve is criteria mapping. Foreign criteria are namespaced, but can be mapped into strict equivalences with local criteria (`implies` but in both directions). The two builtins should automap, but the user can also specify custom M:N mappings. How to apply that mapping efficiently is a bit of a concern for me. As far as I can tell, you need to do a quadratic amount of work in the worst-case to apply the mappings.

Consider the case of only 2:1 mappings, (`[x]` is set, `[ ]` is unset):

```
mapping 1: [ ][ ][ ][x][x] [ ][ ][ ][ ][ ] == [ ][ ][ ][ ][ ] [x][ ][ ][ ][ ]
mapping 2: [ ][ ][x][ ][ ] [x][ ][ ][ ][ ] == [ ][ ][ ][ ][ ] [ ][x][ ][ ][ ]
mapping 3: [ ][x][ ][ ][ ] [ ][x][ ][ ][ ] == [ ][ ][ ][ ][ ] [ ][ ][x][ ][ ]
mapping 4: [x][ ][ ][ ][ ] [ ][ ][x][ ][ ] == [ ][ ][ ][ ][ ] [ ][ ][ ][x][ ]
```

Then with this input, we have a cascading series of equalities:

```
[x][x][x][x][x] [ ][ ][ ][ ][ ]
[x][x][x][x][x] [x][ ][ ][ ][ ]  via mapping 1
[x][x][x][x][x] [x][x][ ][ ][ ]  via mapping 2
[x][x][x][x][x] [x][x][x][ ][ ]  via mapping 3
[x][x][x][x][x] [x][x][x][x][ ]  via mapping 4
```

At each step we have one criteria that wasn't involved before, and one criteria that only came into existence in the last step. I believe that this shows that you cannot simply do as we currently do with "implies" where you compute some kind of transitive closure ahead of time. Nor can you do a random-access lookup against your currently set criteria for what rules definitely apply.

Seemingly this means that you need to linearly scan over every mapping rule in order and apply them. Then you just need to do it all over again until you hit a fixed-point. Such a process would be `O(k * j)` with `k` criteria and `j` rules. The process will definitely terminate at this point as we are always making progress on each linear scan. Note that this cost would have to be paid for every `union_with` or `set_criteria` we do. So if we do `g` of those, we're talking about `O(k * j * g)`.

One possible way to optimize this would be to have a mapping from individual criteria to all the rules that refer to them. So whenever we apply a rule, we can know what criteria are newly set and only check mappings that involve those criteria. Such a mapping would have `O(k * j)` size worst-case, and therefore take just as long to compute, *but* that would be a fixed cost instead of a cost we pay on every operation. Sadly I think you could still tune such an input such that the each entry has `O(j)` rules so this doesn't gain much in theory. In *practice* it might gain a lot, as most rules will only refer to a few criteria, and we will go from having to naively check every rule to knowing exactly what rules to check.

This is probably overthinking things but the resolver is a combinatoric time bomb, so shaving off multiplicative factors is worth some consideration. Also I have insomnia and apparently that means algorithms brain.

Gankra:
fixes #240 in principle, but needs ALLLL the tests

This ended up being less gnarly than I expected, I especially like that it results in a bunch of duplication getting removed. 

Right now I have duped all the APIs as `criteria_mapper.blah(...)` and `criteria_mapper.namespaced_blah(namespace, ...)`. The non-namespaced version is implicitly Local. It's convenient but a potential footgun. Not sure if it should be removed in favour of all call-sites being explicit (anything from audits.toml and config.toml is always local, except for the foreign mappings themselves).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -111,7 +111,7 @@ dependencies = [
 
 [[package]]
 name = "cargo-vet"
-version = "0.1.0"
+version = "0.2.0"
 dependencies = [
  "cargo_metadata",
  "clap",
--- a/src/main.rs
+++ b/src/main.rs
@@ -740,25 +740,30 @@ fn do_cmd_certify(
             writeln!(out);
             writeln!(out, "  0. <clear selections>");
             let implied_criteria = criteria_mapper.criteria_from_list(&chosen_criteria);
-            for (criteria_idx, (_namespace, criteria_name, _criteria_entry)) in
-                criteria_mapper.list.iter().enumerate()
-            {
-                if chosen_criteria.contains(criteria_name) {
+            for (criteria_idx, criteria_info) in criteria_mapper.list.iter().enumerate() {
+                if chosen_criteria.contains(&criteria_info.namespaced_name) {
                     writeln!(
                         out,
                         "  {}. {}",
                         criteria_idx + 1,
-                        out.style().green().apply_to(criteria_name)
+                        out.style().green().apply_to(&criteria_info.namespaced_name)
                     );
                 } else if implied_criteria.has_criteria(criteria_idx) {
                     writeln!(
                         out,
                         "  {}. {}",
                         criteria_idx + 1,
-                        out.style().yellow().apply_to(criteria_name)
+                        out.style()
+                            .yellow()
+                            .apply_to(&criteria_info.namespaced_name)
                     );
                 } else {
-                    writeln!(out, "  {}. {}", criteria_idx + 1, criteria_name);
+                    writeln!(
+                        out,
+                        "  {}. {}",
+                        criteria_idx + 1,
+                        &criteria_info.namespaced_name
+                    );
                 }
             }
 
@@ -798,7 +803,7 @@ fn do_cmd_certify(
                 writeln!(out, "error: not a valid criteria");
                 continue;
             }
-            chosen_criteria.push(criteria_mapper.list[answer - 1].1.clone());
+            chosen_criteria.push(criteria_mapper.list[answer - 1].namespaced_name.clone());
         }
         chosen_criteria
     } else {
--- a/src/main.rs
+++ b/src/main.rs
@@ -705,7 +705,12 @@ fn do_cmd_certify(
         (user_info.username, Some(who))
     };
 
-    let criteria_mapper = CriteriaMapper::new(&store.audits.criteria);
+    //let foreign_criteria = store.imports.iter().map();
+    let criteria_mapper = CriteriaMapper::new(
+        &store.audits.criteria,
+        &store.imports,
+        &store.config.imports,
+    );
 
     let criteria_names = if sub_args.criteria.is_empty() {
         let (from, to) = match &kind {
@@ -735,7 +740,7 @@ fn do_cmd_certify(
             writeln!(out);
             writeln!(out, "  0. <clear selections>");
             let implied_criteria = criteria_mapper.criteria_from_list(&chosen_criteria);
-            for (criteria_idx, (criteria_name, _criteria_entry)) in
+            for (criteria_idx, (_namespace, criteria_name, _criteria_entry)) in
                 criteria_mapper.list.iter().enumerate()
             {
                 if chosen_criteria.contains(criteria_name) {
@@ -793,7 +798,7 @@ fn do_cmd_certify(
                 writeln!(out, "error: not a valid criteria");
                 continue;
             }
-            chosen_criteria.push(criteria_mapper.list[answer - 1].0.clone());
+            chosen_criteria.push(criteria_mapper.list[answer - 1].1.clone());
         }
         chosen_criteria
     } else {
--- a/src/main.rs
+++ b/src/main.rs
@@ -31,7 +31,9 @@ use crate::format::{
     SortedMap, StoreInfo,
 };
 use crate::out::Out;
-use crate::resolver::{Conclusion, CriteriaMapper, DepGraph, ResolveDepth, SuggestItem};
+use crate::resolver::{
+    Conclusion, CriteriaMapper, CriteriaNamespace, DepGraph, ResolveDepth, SuggestItem,
+};
 use crate::storage::{Cache, Store};
 
 mod cli;
@@ -705,7 +707,6 @@ fn do_cmd_certify(
         (user_info.username, Some(who))
     };
 
-    //let foreign_criteria = store.imports.iter().map();
     let criteria_mapper = CriteriaMapper::new(
         &store.audits.criteria,
         &store.imports,
@@ -740,7 +741,15 @@ fn do_cmd_certify(
             writeln!(out);
             writeln!(out, "  0. <clear selections>");
             let implied_criteria = criteria_mapper.criteria_from_list(&chosen_criteria);
-            for (criteria_idx, criteria_info) in criteria_mapper.list.iter().enumerate() {
+            // Iterate over all the local criteria. Note that it's fine for us to do the enumerate
+            // first, because local criteria are added to the list before foreign criteria, so they
+            // should be contiguous from 0..N.
+            let local_criteria = criteria_mapper
+                .list
+                .iter()
+                .enumerate()
+                .filter(|(_, info)| matches!(info.namespace, CriteriaNamespace::Local));
+            for (criteria_idx, criteria_info) in local_criteria {
                 if chosen_criteria.contains(&criteria_info.namespaced_name) {
                     writeln!(
                         out,
--- a/src/resolver.rs
+++ b/src/resolver.rs
@@ -73,7 +73,7 @@ use tracing::{error, trace, trace_span};
 use crate::errors::SuggestError;
 use crate::format::{
     self, AuditKind, CriteriaName, CriteriaStr, Delta, DiffStat, ExemptedDependency, ImportName,
-    ImportsFile, PackageName, PackageStr, PolicyEntry, RemoteImport,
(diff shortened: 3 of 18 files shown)


src/resolver.rs
```diff
@@ -472,7 +581,7 @@ impl CriteriaMapper {
                     cur_idx == other_idx || !self.implied_criteria[other_idx].has_criteria(cur_idx)
                 })
             })
-            .map(|idx| &*self.list[idx].0)
+            .map(|idx| &*self.list[idx].1)
```
Gankra, line 584: Note to self, in principle this should get some kind of disambiguating prefix like `format("{.0}::{.1}")` in the case of foreign criteria. I'm actually inclined to implement this as inserting that preformatted string into `self.list[..].1` itself, and making other code `split_once`. That or we add two strings to the list, one for the "local name" and one for the "global name". Now that I'm saying it that sounds a lot simpler lol.

Gankra: Oh other note to self, this code is completely busted. The `!implied.has_criteria` check is fundamentally assuming non-circularity. We should tiebreak circularity by bypassing this guard when cur_idx is `Local` and other_idx` is `Foreign`.

Gankra: Final note to self, different code probably wants different behaviour on showing/hiding foreign criteria. But needs experimentation.

src/tests/snapshots/cargo_vet__tests__certify__mock-simple-certify-flow.snap
```diff
@@ -6,12 +6,12 @@ OUTPUT:
...
-  3. strong-reviewed
-  4. weak-reviewed
-  5. safe-to-deploy
-  6. safe-to-run
+  1. safe-to-run
+  2. safe-to-deploy
+  3. fuzzed
+  4. reviewed
+  5. strong-reviewed
+  6. weak-reviewed
```
Gankra, line 14: This inversion was intentional, it feels Morally Right that builtins should be first.

src/resolver.rs
```diff
@@ -368,46 +370,123 @@ fn builtin_criteria() -> SortedMap<CriteriaName, CriteriaEntry> {
...
+            let prev = index
+                .entry(namespace.clone())
+                .or_default()
+                .insert(criteria_name.clone(), idx);
+            assert!(prev.is_none(), "criteria name was multiply defined???");
+        }
+
+        // Hack the builtins into every foreign namespace. Because they all use
+        // the same CriteriaIdx, we can now forget that builtins are special and
+        // just naturally look them up in any namespace without issue.
```
Gankra, line 418: Hmm this might not work quite right when using namespace as AuditSource when e.g. handling blames? I think it works out but needs consideration.

Cargo.lock
```diff
@@ -111,7 +111,7 @@ dependencies = [
 
 [[package]]
 name = "cargo-vet"
-version = "0.1.0"
+version = "0.2.0"
```
mystor, line 114: I'm a bit surprised our `Cargo.lock` wasn't already updated. I guess we hadn't run tests since the version bump?

src/resolver.rs
```diff
@@ -188,14 +181,28 @@ pub struct CriteriaFailureSet {
...
+    Local,
+    Foreign(ImportName),
+}
+
+#[derive(Debug, Clone)]
+pub struct CriteriaInfo {
+    pub namespace: CriteriaNamespace,
+    pub raw_name: CriteriaName,
+    pub namespaced_name: String,
+    pub entry: CriteriaEntry,
```
mystor, line 195: It seems like it might be nice to include the implies CriteriaSet in here too, rather than just holding the single CriteriaEntry. We could have a `direct_implies` set which is populated immediately from the entries implies relationship, and then fill in `implies` afterward recursively like we're already doing.

I don't know if we really want the full CriteriaEntry here because the only extra info it has over that right now is descriptions, which could be large and IIRC the resolver never uses internally so aren't important to hold onto (though I could be forgetting something there).

src/resolver.rs
```diff
@@ -188,14 +181,28 @@ pub struct CriteriaFailureSet {
...
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash, Serialize)]
+pub enum CriteriaNamespace {
+    Local,
+    Foreign(ImportName),
+}
+
+#[derive(Debug, Clone)]
+pub struct CriteriaInfo {
+    pub namespace: CriteriaNamespace,
+    pub raw_name: CriteriaName,
```
mystor, line 193: It looks like this is only used when building up the mapper and isn't accessed after that, it might be nice to not hold onto the raw name if we don't need it.

The `namespace` is also used when handling cross-namespace implies cycles, though if we dedupe at the index level when doing mappings, we could also probably remove that field, as it's not accessed other than there and when building up the mapper (which I believe could be done in a single pass).

mystor (commented):
There are some implementation strategy comments here, but mostly I'm hoping to double-check thinking on some things and make sure the behaviour changes which were made are intentional.

src/main.rs
```diff
@@ -735,25 +740,30 @@ fn do_cmd_certify(
...
-                if chosen_criteria.contains(criteria_name) {
+            for (criteria_idx, criteria_info) in criteria_mapper.list.iter().enumerate() {
+                if chosen_criteria.contains(&criteria_info.namespaced_name) {
                     writeln!(
                         out,
                         "  {}. {}",
                         criteria_idx + 1,
-                        out.style().green().apply_to(criteria_name)
+                        out.style().green().apply_to(&criteria_info.namespaced_name)
                     );
```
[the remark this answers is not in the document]
Gankra: key insights:

* the builtin criteria *are* deduped, so in the common case there are no namespaces
* when sugarring a CriteriaSet to CriteriaNames, the already have deduping logic that does "safe-to-run, safe-to-deploy" => "safe-to-deploy". this logic has been updates to similarly do "X, foreign::X" => "X". So we will only ever report "foreign::X" when it's not mapped in, and therefore there's nothing else to refer to it by.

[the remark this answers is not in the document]
Gankra: btw you *can't* certify someone else's criteria, so there is definitely some dubious UX here. I'm currently naively hoping that the user will see that, go "hey what the hell", and subsequently do some manual reasoning about how to resolve the fact that we failed to audit something due to missing foreign audits.

src/resolver.rs
```diff
@@ -500,12 +629,19 @@ impl CriteriaMapper {
...
+                        // Ignore our own index
+                        let is_identity = cur_idx == other_idx;
+                        // Discard this criteria if it's implied by another
+                        let isnt_implied = !self.implied_criteria[other_idx].has_criteria(cur_idx);
+                        // Unless we're local and they're foreign, then we win
+                        let cur_is_local = self.list[cur_idx].namespace == CriteriaNamespace::Local;
+                        let other_is_foreign =
+                            self.list[other_idx].namespace != CriteriaNamespace::Local;
+                        let is_mapping = cur_is_local && other_is_foreign;
+                        is_identity || isnt_implied || is_mapping
```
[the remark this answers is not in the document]
Gankra: fair

src/resolver.rs
```diff
@@ -1351,75 +1491,28 @@ fn resolve_third_party<'a>(
     let package = &graph.nodes[pkgidx];
     let exemptions = store.config.exemptions.get(package.name);
 
-    let own_audits = store.audits.audits.get(package.name).unwrap_or(&NO_AUDITS);
+    let own_audits = (
+        CriteriaNamespace::Local,
+        store.audits.audits.get(package.name).unwrap_or(&NO_AUDITS),
+    );
+    // FIXME: ideally we wouldn't clone the CriteriaNamespace every time we iterate all_criteria
+    // (which happens a lot when inspecting violations).
```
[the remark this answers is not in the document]
Gankra: what I really wanted is to use Borrow here and have CriteriaNamespaceStr, but I don't think you can unsize enums and Borrow requires the reference to be on the outside😿

Cargo.lock
Gankra: yeah this through me off, i guess CI is locked so it doesn't notice

src/resolver.rs
Gankra: this is a good point. i think holding onto the entry was largely an artifact of the long-past days where CriteriaMapper tried really hard to not do any clones and just hold onto borrowed things, but builtins got in the way of this, so that hasn't been the case for a while.

src/resolver.rs
Gankra: it's also used to do the latebound deduping when getting a list of set criteria, but yeah if we did early-bound deduping we might not need it..?

src/main.rs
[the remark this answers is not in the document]
mystor: The problem here is that this is a UI where you can enter numbers to choose which criteria to certify for, and us listing a foreign criteria here definitely implies that you can select it and certify for it!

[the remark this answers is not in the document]
Gankra: oh sorry yes in this precise area only local names should show up (but I used namespaced_name to make it more obvious if we fuck up). I should double check that I'm not leaking them in (probably am).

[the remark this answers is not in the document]
mystor: I think right now we unconditionally include every mapped criteria, but I could be misremembering.

Issue #278, opened 2022-07-21
"uncomment the statement below" is ambiguous


bholley:
@gw3583 found the current setup confusing, in that he thought he was supposed to uncomment the criteria as well.

I wonder if we should consider replacing it with "and then follow the instructions below", and the replacement "STATEMENT" with "uncomment the following statement:" and "NOTES" with "Add any notes about your audit below this line".

PR #280, opened 2022-07-22
Certify Flow UX Improvements

### LINKED ISSUE #277: Instruct the user about |cargo vet certify| after launching inspect
Right now `cargo vet suggest` includes the following message:

> Use |cargo vet certify| to record the audits.

However, once somebody does a `cargo vet inspect`, their terminal has been clobbered already and the next step is no longer obvious.

My put here is that after the user hits enter (to open the in-browser source inspection), we should print something like the above message again, so that when they come back to their terminal it's clear what the next step is.

It's a little less clear to me how to optimally handle this with the subshell in the `--local` case, but we can punt on that for now if needed.

mystor:
A set of wording changes to the certify flow in order to improve their clarity. The bulk of the changes here are to make sure that we release the cache and store locks before running a subshell or showing diffs when running in local mode.

This change ends up removing `ExecPanic` as we need to keep the current process around to print a message after the shell has exited.

Fixes #277 and #278.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -838,7 +838,7 @@ fn do_cmd_certify(
         }
 
         editor.add_comments(
-            "Please read the following criteria and uncomment the statement below:",
+            "Please read the following criteria and then follow the instructions below:",
         )?;
         editor.add_text("")?;
 
@@ -850,11 +850,11 @@ fn do_cmd_certify(
             editor.add_comments("=== END CRITERIA ===")?;
             editor.add_comments("")?;
         }
-        editor.add_comments("STATEMENT:")?;
+        editor.add_comments("Uncomment the following statement:")?;
         editor.add_text("")?;
         editor.add_comments(&statement)?;
         editor.add_text("")?;
-        editor.add_comments("NOTES:")?;
+        editor.add_comments("Add any notes about your audit below this line:")?;
         editor.add_text("")?;
         if let Some(notes) = &notes {
             editor.add_text(notes)?;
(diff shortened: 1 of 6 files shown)


src/main.rs
```diff
@@ -1521,6 +1524,8 @@ fn cmd_diff(out: &Arc<dyn Out>, cfg: &Config, sub_args: &DiffArgs) -> Result<(),
         .map_err(CommandError::CommandFailed)
         .into_diagnostic()?;
 
+    writeln!(out, "\nUse |cargo vet certify| to record your audits.");
```
bholley, line 1527: Nit: I'd make "audits" singular in this and the other instances.

src/main.rs
```diff
@@ -109,10 +109,6 @@ const DURATION_DAY: Duration = Duration::from_secs(60 * 60 * 24);
 /// by panicking with this type instead of a string.
 struct ExitPanic(i32);
 
-/// Similar to the above, but allows us to exec a new command
-/// as our final act.
-struct ExecPanic(std::process::Command);
```
Gankra, line 114: 🎊 this is good motivation for me to remove the last of my error handling sins and remove the Last Remaining ExitPanic.

src/main.rs
```diff
@@ -530,75 +518,85 @@ fn cmd_inspect(
...
+            .into_diagnostic()?;
 
-    #[cfg(not(target_family = "unix"))]
-    {
-        writeln!(out, "  fetched to {:#?}", fetched);
-        Ok(())
+        writeln!(out, "\nUse |cargo vet certify| to record your audit.");
+
+        if let Some(code) = status.code() {
+            panic_any(ExitPanic(code));
```
Gankra, line 592: oh rip lol. we'll get 'em eventually.

I'm assuming I can do something like have a ExitError and dynamic cast it out of the Report...

Gankra (approved):

PR #296, opened 2022-08-19
Update imports.lock less frequently

### LINKED ISSUE #272: Imported audits may change `imports.lock` each time `cargo vet` is run
After a discussion with Wasmtime folks and @bholley yesterday about using `cargo vet` in the Bytecode Alliance, specifically the Wasmtime project, it sounded like we would likely want to get into a situation where audits are imported from Firefox (and perhaps vice-versa one day). In doing so though it sounded like each time `cargo vet` is run locally by Wasmtime contributors that the `imports.lock` would continue to grow with the audits added to Firefox.

One situation that we were worried about was where a contributor sends a PR with changes to `imports.lock`, presumably imported from Firefox. When checking this locally to verify that `imports.lock` doesn't change for us it sounded like `imports.lock` could indeed change if Firefox added more audits relative to when the contributor themsleves ran `cargo vet`. I believe this was because `imports.lock` largely copied the `audits.toml` from imported files without filter for what was actually used. This felt surprising because we would not be able to guarantee locally that running `cargo vet` wouldn't actually change the `imports.lock` file if the contributor had correctly run `cargo vet` to import an audit from Firefox.

Two possible ideas we thought might mitigate this were:

* Fetching the most recent contents of `audits.toml` from improted sources could be delayed until there's an actual audit failure. This still sort of runs the issue of something needs to check that `imports.lock` is indeed contained in the remote `audits.toml`
* The contents of `imports.lock` could be filtered to a subset of imported audits which apply to the crates in the local dependency graph. This way so long as Firefox doesn't remove audits for example no matter what new crates were added we'd still generate the same `imports.lock` locally.

mystor:
This patch reduces the frequency that imports.lock updates by doing a
few things:

 * Imports are only updated when necessary for the vet to pass. If vet
   would successfully pass without the new audits, only changes to
   notes, criteria descriptions, revoked audits, and violations will be
   pulled.

 * Some commands, such as suggest, will always run against the live
   copy when not --locked, providing the most accurate results possible.

 * Only audits for crates in your crate graph will be imported.

 * When an update is required, all new audits for the given crate will
   be fetched and locked, to batch updates as much as possible.

This required doing some changes to how imports are handled under the
hood, as well as a number of changes to the resolver in order to track
whether fresh imports are required with sufficient accuracy. Imports are
also now fetched in more APIs in order to have the most up-to-date
information.

The changes to the resolver involved piping through "caveats" which are
determined when solving paths. This is an expansion on the existing
"needs_exemption"/"fully_audited" flags which were being used to
generate stats after a successful vet to also handle freshly imported
audits which should only be used when necessary.

Special handling was also needed to make sure that audits for unused
criteria are not counted by tracking caveats on a per-criteria basis and
bubbling them out.

Fixes #272

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -132,6 +132,7 @@ dependencies = [
  "reqwest",
  "serde",
  "serde_json",
+ "similar",
  "tar",
  "tempfile",
  "textwrap",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -57,3 +57,4 @@ features = [
 
 [dev-dependencies]
 insta = "1.15.0"
+similar = "2.1.0"
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -185,6 +185,12 @@ pub enum StoreAcquireError {
     #[error(transparent)]
     #[diagnostic(transparent)]
     Validate(#[from] StoreValidateErrors),
+    #[error(transparent)]
+    #[diagnostic(transparent)]
+    FetchAuditError(#[from] FetchAuditError),
+    #[diagnostic(transparent)]
+    #[error(transparent)]
+    CriteriaChange(#[from] CriteriaChangeErrors),
 }
 
 #[derive(Debug, Error, Diagnostic)]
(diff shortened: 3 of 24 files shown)


bholley:
Gankra gave this the thumbs-up.

PR #298, opened 2022-08-19
Add 6000 words of documentation on The Algorithm

Gankra:


Gankra:
(Haven't actually tried building this, sleep time)

book/src/algorithm.md
```diff
@@ -0,0 +1,790 @@
...
+
+A lot of the heavy lifting for this task is in Step 1c (AuditGraph).
+
+Trying to validate all criteria at once is slightly brain-melty (because
+different criteria may be validated by different paths), so as a simplifying
+step we validate each criteria individually (so everything I'm about to
+describe happens in a for loop).
+
+If all we care about is finding out if a package has some criteria, then all
+we need to do is run depth-first-search ([DFS][]) from the Root Node and see if it reaches
```
mystor, line 493: I'm not sure if the code is technically a depth-first search anymore given that it's now using a priority queue rather than a stack (since #296). Heaps don't preserve node order in the same way that a stack would.

I suppose it's more of a [Best First Search](https://en.wikipedia.org/wiki/Best-first_search) but with the edge weights simplified because no successor can have better hereustics than its parent, and the hereustics have nothing to do with distance to target (though that is a potential thing we could do, by adding a distance into the sorting key)

book/src/algorithm.md
```diff
@@ -0,0 +1,790 @@
...
+to allow a self_policy to be either *weaker* or *stronger* than the requirements
+of the reverse-dependency.
+
+To further indicate this, when checking a self_policy we either set the node's
+validated criteria to all_criteria or no_criteria.
+
+It's also worth noting that, due to audit-as-crates-io (and maybe other shenanigans
+like patches?) you can end up in a situation where a third-party depends on a
+first-party, or third-parties have `policy.criteria` entries. This is why
+Step 3a is interleaved with Step 2.
```
mystor, line 614: Yup, `[patch]` can also do this.

mystor:
The test failure is existing, and I've put up a fix in #301

PR #299, opened 2022-08-19
Generate json snapshots in all snapshot tests

mystor:
Previously these tests would only generate snapshots for the human
output of cargo vet. With this change, json snapshots are also produced.
This both acts as a test for the json output from cargo vet, as well as
a way to see more information about the vet's results, as json output is
occasionally more detailed than human output.

Finally, it also acts as a source of examples for the JSON output.


mystor:
Ah, the tests are failing because I generated the snapshots on top of #296, oops.

PR #302, opened 2022-08-24
Improve the precision and exemption re-use of "regenerate exemptions"

### LINKED ISSUE #262: Certify after a version bump should remove a pre-existing exemption
Ran into this while updating the version of a crate that had never been audited before. I performed the version bump, certified the new version, and then had to manually remove the exemption from config.toml. I was told by @bholley that `cargo vet` should attempt to remove an exemption in this case and is likely being confused by the version bump.

STR:
```bash
mkdir test
cd test

cargo init --library
# add wast 43.0.0
vim Cargo.toml
cargo vet init
# note exemptions.wast exists in config.toml

vim Cargo.toml # bump wast to 44.0.0
cargo update
cargo vet # fails!

cargo vet certify wast 44.0.0
cargo vet # succeeds!

# note exemptions.wast 43.0.0 still exists in config.toml
```

### LINKED ISSUE #292: `regenerate exemptions` is too aggressive
In mozilla-central today, we an exemption for packed_simd_2 version 0.3.7 and then a delta audit from 0.3.7 to 0.3.8. The idea of this is to pin our exemptions in the past and audit everything new, so that we are protected from future supply-chain attacks even if we haven't gotten around to auditing all the pre-existing dependencies.

At present, if you run  `cargo vet regenerate exemptions`, the exemption will be updated to 0.3.7, which is not what we want. We had originally [discussed](https://github.com/mozilla/cargo-vet/issues/33#issuecomment-1138878109) an approach to avoid this from happening, but unclear if that didn't get implemented or if it did and this is just a bug.

mystor:
This patch reworks the `minimize_exemptions` logic to be more precise
and fix issues such as those found in #292.

This new algorithm doesn't use the `suggest` backend, instead
iteratively selecting exemptions which are required to allow resolve to
proceed, and then re-trying the resolver as-necessary. This is an
improvement over the existing logic as it is able to be much more
precise while prioritizing keeping exemptions pinned in the past over
crate sizes.

The algorithm also has logic to attempt to preserve some special
exemptions, such as those with `dependency_criteria` or which have been
marked `suggest = false`, so that they are not lost when regenerating,
while not allowing the criteria they're exempting to expand (as that
would be potentially surprising for both cases).

If there are no existing audits for a crate which can be used as a
guideline for a newly-required audit, the new algorithm will always bias
toward doing a full-exemption for the current crate version, over trying
to pick an arbitrary node to audit, as this simplifies the logic
substantially and should produce acceptable results for a newly added
dependency subtree.

Unlike the previous approach, there should be no cases where `vet` fails
after a successful `cargo vet regenerate`, as it correctly handles
`dependency_criteria` due to the iterative resolver calls.

If we decide to eventually run this automatically when a `vet` is
successful (e.g. in `cargo vet certify` to fix #262), it would be fairly
straightforward to extend this with an option to prevent the possibility
of new exemptions being added.

Fixes #292

### DIFF
--- a/book/src/config.md
+++ b/book/src/config.md
@@ -170,6 +170,21 @@ Specifies the exact version which should be exempted.
 
 Specifies the criteria covered by the exemption.
 
+#### `dependency-criteria`
+
+Allows overriding the criteria requirements for dependencies on a per-dependency basis.
+Similar in format to the [equivalent field](audit-entries.md#dependency-criteria) in audit entries.
+
+This serves the same purposes as the field on audit entries, allowing the
+exemption to relax or strengthen the requirements which it places on
+dependencies when it is used.
+
+This can be used when a crate still needs to be exempted (e.g. because it hasn't
+been audited enough to publish an audit), but it has been determined that a
+particular subtree should be held to different audit requirements. This may both
+be useful for dependencies which only need to be `safe-to-run`, or for adding
+extra requirements for specific dependencies of an exempted crate.
+
 #### `notes`
 
 Free-form string for recording rationale or other relevant information.
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -135,17 +135,16 @@ pub enum InitError {
 }
 
 //////////////////////////////////////////////////////////
-// MinimizeUnauditedError
+// RegenerateExemptionsError
 //////////////////////////////////////////////////////////
 
 #[derive(Debug, Error, Diagnostic)]
 #[non_exhaustive]
-pub enum MinimizeUnauditedError {
-    #[error(transparent)]
-    #[diagnostic(transparent)]
-    Suggest(#[from] SuggestError),
-    #[error("An unknown error occured while trying to minimize exemptions entries")]
-    Unknown,
+pub enum RegenerateExemptionsError {
+    #[error(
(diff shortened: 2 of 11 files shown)


book/src/config.md
```diff
@@ -170,6 +170,17 @@ Specifies the exact version which should be exempted.
 
 Specifies the criteria covered by the exemption.
 
+#### `dependency-criteria`
+
+Allows overriding the criteria requirements for dependencies on a per-dependency
+basis.  Allows overriding the above values on a per-dependency basis. Similar in
```
bholley, line 176: Are these sentences redundant?

book/src/config.md
```diff
@@ -170,6 +170,17 @@ Specifies the exact version which should be exempted.
...
+format to the [equivalent field](audit-entries.md#dependency-criteria) in audit
+entries.
+
+This serves the same purposes as the field on audit entries, allowing the
+exemption to relax or strengthen the requirements which it places on
+dependencies when it is used.
```
bholley, line 182: Maybe get more specific about the use-case: you might not have fully audited some crate but you can at least tell that a subtree can be mostly ignored.

bholley (approved):

book/src/config.md
mystor: They sure are! oops

PR #313, opened 2022-08-26
Stop using 0.0.0 as the root version

### LINKED ISSUE #307: Stop using `0.0.0` as the root version
The logic in the resolver currently treats the version `0.0.0` as the "root version" (i.e. equivalent to no version at all). Unfortunately for us, that version is one which can definitely be published (e.g. https://crates.io/crates/cursed-trying-to-break-cargo/0.0.0), meaning that we'll misbehave when trying to vet version `0.0.0` of a crate, as it will always be considered fully audited as the root version.

We should update the logic in the resolver to operate on `Option<Version>` instead of `Version` in places which need a "root version" going forward.

mystor:
The logic in the resolver previously treated the version 0.0.0 as the
"root version" (i.e. equivalent to no version at all). Unfortunatley,
this version can be published, meaning that we'll misbehave when trying
to interact with an actually published version 0.0.0 of a crate, and
will always consider it fully-audited as the "root version".

This patch updates the logic in the resolver to operate on
`Option<Version>` instead of `Version` in places where the root version
is valid.

Because this encoding of 0.0.0 as the root version is saved on disk in
the diff-cache, this patch changes the diff-cache format in an
incompatible way to force it to be re-built. It also adds a mechanism to
make forcing the diff-cache to be re-built easier in the future.

Unfortunately, these changes lead to a large number of JSON output test
failures, as we exposed the old internal desugaring of '0.0.0' as a diff
source in the JSON output, which has now changed to null. We may want to
look into changing this part of the output to avoid talking about full
audits as "diffs from null" in the future.

Fixes #307

### DIFF
--- a/book/src/algorithm.md
+++ b/book/src/algorithm.md
@@ -360,9 +360,8 @@ Each edge has a list of criteria is claims to certify, and dependency_criteria t
 dependencies of this package must satisfy for the edge to be considered "valid" (see
 the next section for details).
 
-There is an implicit Root Version which represents an empty package. As of this writing
-the Root Version is simply 0.0.0, but this isn't really correct and nodes should be more
-like `Option<Version>`.
+There is an implicit Root Version which represents an empty package, meaning that throughout
+much of the audit graph, versions are represented as `Option<Version>`.
 
 When trying to validate whether a particular version of a package is audited, we also add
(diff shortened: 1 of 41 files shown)


Gankra (approved):
dope, thanks for fixing my mistakes!

PR #315, opened 2022-08-26
Allow policy dependency_criteria for crates-io crates

mystor:
Previously this wouldn't be allowed, as the dependency-criteria would be
pulled from the audit edges, rather than from the policy. With these
changes, it will be pulled from both places, with policy
dependency-criteria overriding the criteria specified by all audit and
exemption edges.

### DIFF
--- a/book/src/config.md
+++ b/book/src/config.md
@@ -116,10 +116,10 @@ of a given crate.
 
 ### the `policy` Table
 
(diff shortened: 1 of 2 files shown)


mystor:
cc @bholley I'm not sure if we want to do this yet, but it's a rough implementation (without tests) of the idea we briefly discussed on matrix.

PR #316, opened 2022-08-27
TEST: update to clap 4.0.0-prerelease

Gankra:
this is a demo for https://github.com/clap-rs/clap/issues/4132

### DIFF
--- a/tests/snapshots/test_cli__long-help.snap
+++ b/tests/snapshots/test_cli__long-help.snap
@@ -3,17 +3,47 @@ source: tests/test-cli.rs
 expression: format_outputs(&output)
 ---
 stdout:
-cargo-vet 0.3.0
+cargo-vet-vet 0.3.0
 Supply-chain security for Rust
 
 When run without a subcommand, `cargo vet` will invoke the `check` subcommand. See `cargo vet help
 check` for more details.
 
-USAGE:
+Usage:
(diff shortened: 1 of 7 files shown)


tests/snapshots/test_cli__long-help.snap
```diff
@@ -3,17 +3,47 @@ source: tests/test-cli.rs
 expression: format_outputs(&output)
 ---
 stdout:
-cargo-vet 0.3.0
+cargo-vet-vet 0.3.0
```
Gankra, line 6: This is a regression, unclear whose fault it is (see https://github.com/clap-rs/clap/issues/4132#issuecomment-1229101979).

epage: If you set `name = cargo` in addition to the bin_name, this should be resolved.  See https://github.com/clap-rs/clap/blob/master/examples/cargo-example-derive.rs#L4

In clap 3, the user sets `bin_name = cargo` and we build that up as we recurse down the subcommands.  When we then render the "name", we then did a search/replace on it which on windows without an explicit `bin_name` would end up with `foo.exe-bar`.  So in clap 4, we build up the `usage_name` from the root command's `bin_name` + each subcommand's `name` and the `display_name` from all command's `name`s.

PR #323, opened 2022-09-07
Implement cargo vet aggregate

### LINKED ISSUE #83: Implement `cargo vet aggregate`
The current docs recommend handling multiple repositories with either star-topological imports or certifying audits in a dedicated repo. Neither of these is entirely ideal: the former is unwieldy to configure, and the latter is unwieldy to use.

A better solution is to have a single bot-managed repository aggregate audits from all the individual repositories into a single combined record. This enables two things:
1. Each of the individual N repositories can then import this single aggregated record rather than directly referencing all N-1 other repositories.
2. External consumers don't need to concern themselves with the repository details and can just do a single import for that organization.

I think the smoothest way to do this would be to implement a `cargo vet aggregate` subcommand which knows how to fetch and merge a list of audit file URLs. We could then include instructions in the doc for setting up a GitHub repo with a cron-driven Action to poll the relevant URLs every five minutes and commit any updates.

mystor:
This feature allows specifying multiple sources in a `sources.list` file which will be processed to generate a new source file, containing all of the audits from each listed source. Criteria are merged from these sources by-name, with the descriptions and implies sets requried to be identical between criteria with the same name from different sources.

The chain of what imports were used for a given criteria or audit is now tracked in the .toml file as well, using an aggregated-from field on the relevant entries.

Fixes #83

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -322,6 +322,15 @@ pub enum Commands {
     #[clap(disable_version_flag = true)]
     FetchImports(FetchImportsArgs),
 
+    /// Fetch and merge audits from multiple sources into a single `audits.toml`
+    /// file.
+    ///
+    /// Will fetch the audits from each URL in the provided file, combining them
+    /// into a single file. Custom criteria will be merged by-name, and must
+    /// have identical descriptions in each source audit file.
+    #[clap(disable_version_flag = true)]
+    Aggregate(AggregateArgs),
+
     /// Print the cargo build graph as understood by `cargo vet`
     ///
     /// This is a debugging command, the output's format is not guaranteed.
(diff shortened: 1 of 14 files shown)


mystor:
For an example of what aggregated sources look like in practical usage, this is an aggregated output of wasmtime, embark and mozilla's current audit files: https://gist.github.com/mystor/f5f42cf940235922466f396aeb09cceb

bholley:
Can you do the string-or-vec trick for aggregated-from as well? LGTM otherwise.

Once this lands, let's set up a supply-chain repo under the mozilla GH org that aggregates mozilla-central and cargo-vet itself.

mystor:
> Can you do the string-or-vec trick for aggregated-from as well? LGTM otherwise.

I intentionally didn't because it seemed like it'd be more clear to folks that there can be multiple entries there if we use an array always, but I totally can.

> Once this lands, let's set up a supply-chain repo under the mozilla GH org that aggregates mozilla-central and cargo-vet itself.

We don't actually use cargo-vet on cargo-vet, I suppose we should look into doing that as well first?

bholley:
> > Can you do the string-or-vec trick for aggregated-from as well? LGTM otherwise.
> 
> I intentionally didn't because it seemed like it'd be more clear to folks that there can be multiple entries there if we use an array always, but I totally can.

I think it's most consistent with what we do elsewhere (elide arrays when they're uncommon).
 
> We don't actually use cargo-vet on cargo-vet, I suppose we should look into doing that as well first?

Yeah. I haven't been in a huge hurry because the primary way we're dogfooding is on m-c, but this gives us a way to dogfood `aggregate`.

PR #340, opened 2022-10-14
Update crates.io index when necessary to avoid errors

mystor:
If we would generate an error due to an `audit-as-crates-io` policy being specified for a crate which doesn't appear to be in the index, it may be because the local copy of the index is out-of-date.

With these changes, if we notice the index is out of date in this way, we'll update our local copy of the index and re-run our checks. This should avoid potential issues with new audits being added for `audit-as-crates-io = true` crates reporting errors on machines which have an out-of-date copy of the index.

In the future, we may want to switch to instead using the sparse http API for crates.io, rather than using `crates-index`, in order to handle running on machines using the new sparse crates.io registry support which may be stabilized soon (https://github.com/rust-lang/cargo/pull/11224). Alternatively it may be worth asking `cargo` to add a new subcommand to check if a specific crate exists in the index, which could take advantage of whatever representation is currently being used as well as any caching cargo performs under the hood.

The bulk of this patch involves changing how the mock index is handled in tests to make it possible to test this updating behaviour.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -1342,7 +1342,8 @@ fn fix_audit_as(cfg: &Config, store: &mut Store) -> Result<(), CacheAcquireError
     // NOTE: In the future this might require Network, but for now `cargo metadata` is a precondition
     // and guarantees a fully populated and up to date index, so we can just rely on that and know
     // this is Networkless.
-    let issues = check_audit_as_crates_io(cfg, store);
+    let mut cache = Cache::acquire(cfg)?;
+    let issues = check_audit_as_crates_io(cfg, store, &mut cache);
     if let Err(AuditAsErrors { errors }) = issues {
         for error in errors {
             match error {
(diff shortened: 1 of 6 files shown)


src/main.rs
```diff
@@ -2027,62 +2027,78 @@ fn first_party_packages_strict<'a>(
...
-                    needs_audit_as_entry.push(NeedsAuditAsError {
-                        package: package.name.clone(),
-                        version: package.version.clone(),
-                    });
+    // If we don't have a registry, we can't check audit-as-crates-io.
+    if !cache.has_registry() {
+        return Ok(());
+    }
+
+    'retry: loop {
```
mystor, line 2039: This is a bit gross, but adding this extra loop was the easiest way to make sure we re-check everything, as we'd otherwise get inconsistent results.

bholley: Is it feasible to factor this into a helper or two rather than adding yet another jump label?

mystor: I ended up moving the cache acquire out and writing it as a recursive pattern, which is I think slightly less gross to read.

bholley: Much nicer, thanks.

PR #347, opened 2022-10-28
Add Zcash audit set to registry

str4d:
This audit set is automatically aggregated from all of our audit sources (currently only one, soon to be two).


bholley:
Thanks!

Issue #330, opened 2022-09-15
github bot workflow


Gankra:
MVP for bot-based UX should be based on github, either on top of dependabot or a different bot. The design should be extensible enough to port to other infra like gitlab or mozilla's stuff, but supporting those is not in scope for the first version. Of course we will keep that future in mind and not wildly hardcode github stuff. Use json messages, standard auth protocols, etc.

Basic workflow sketch:

* cargo-vet fails in ci and spits out json describing the failure (already done)
* this triggers some sort of bot event to have it consume that json and post a comment on the PR saying audits need to be done and linking to a webapp
* the webapp has a ui for displaying what audits needs to be done, and links to the URLs that `cargo vet diff` and friends would use
* the webapp has a ui for specifying you completed the audit and for what criteria, displaying the EULA text for those criteria (and whatever else `certify` supports)
* the webapp has authentication integration with github so when you push certify buttons in the app it does the thing over on github

unclear details:

* do the certifies appear on the existing PR, open new PRs, or push to main immediately?
* are the certify commits expected to be done by the bot account or the auditor's account?
* should there be one commit per audit, or a single "all the audits" commit
  * if the latter, are audits batched up and the user says "all done" in the app, or do they get submitted to github as each one is performed?

Gankra:
#329 is an initial prototype of this but without any authentication logic, so it just has the user copy base64 strings of json blobs around to sheppard data between the app and github

Gankra:
how to test the current impl:

```
git fetch origin pull/329/head:nika
git checkout nika
cd certify-gui; npm install; npm run dev
```

[...]

Issue #312, opened 2022-08-26
Document the JSON output formats


mystor:
Currently, commands can often output JSON data by specifying `--output-format=json` on the command line, however exactly what this should look like for each command hasn't been deeply considered, and the output is somewhat undocumented for the commands which we want to support this on. 

We should consider adding documentation for this information, including perhaps a second `cargo-vet-output-format` crate (name bikeshed-able) which just contains serde declarations for the JSON output formats we want to support serializing & de-serializing, so we can explicitly version and document those separately in always-up-to-date code.

Whatever data format we define should probably also explicitly encode how JSON errors are handled, rather than depending on `miette`'s JSON error format, as that appears to change occasionally (although in what appears to be backwards-compatible ways, see e.g. #290).

---

In terms of which commands we want to support JSON output for, currently we accept the command globally, but IIRC it only impacts the normal output of the following subcommands:

* `cargo vet [check]` - the audit results are encoded as JSON. This is probably the most important API to support JSON.
* `cargo vet suggest` - like `cargo vet check`, this encodes suggestions as JSON.
* `cargo vet dump-graph` - can be used to visualize the crate graph as JSON.

For other API endpoints, the only JSON outputs possible are errors.

PR #350, opened 2022-10-31
Migrate the JSON literal report to proper types

Gankra:
This makes it more plausible for this format to be made available in a subcrate to allow other libraries/tools to easily deserialize it. Also makes it more obvious the places where we're being weirdly inconsistent or exposing random internals because I was lazy/hasty.

A weird side-effect of this is that the order of the fields got shuffled around. It seems the json macro will sort fields by name, while proper derives use the decl order. The latter seems preferred anyway, but unfortunately we take a huge churn to the snapshots, obfuscating that this is a semantic noop.

### DIFF
--- a/src/format.rs
+++ b/src/format.rs
@@ -1,5 +1,6 @@
 //! Details of the file formats used by cargo vet
 
+use crate::resolver::{DiffRecommendation, ViolationConflict};
 use crate::serialization::spanned::Spanned;
 use crate::{flock::Filesystem, serialization};
 use core::{cmp, fmt};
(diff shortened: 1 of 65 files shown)


Gankra:
This is part 1 of the required support of #330 which I wanted to split out because it churns all the test files :(

Gankra:
Added doc-comments to all the new types

src/format.rs
```diff
@@ -532,3 +533,116 @@ pub struct CommandHistory {
...
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct JsonSuggestItem {
+    /// The name of the package
+    pub name: PackageName,
+    /// Any notable parents the package has (can be helpful in giving context to the user)
+    pub notable_parents: String,
+    /// The criteria we recommend auditing the package for
+    pub suggested_criteria: Vec<CriteriaName>,
+    /// The diff (or full version) we recommend auditing
+    pub suggested_diff: DiffRecommendation,
```
mystor, line 635: I realize this is existing, but I don't love how we expose the diff internal representation to the json output. It would be nice if we could have this use a delta-or-version like format like we do for the actual audits.toml format, so that the fact that full diffs are `"null -> 1.0.0"` isn't visible in public APIs.

The raw diffstat being encoded directly in there is also a bit weird.

src/format.rs
```diff
@@ -532,3 +533,116 @@ pub struct CommandHistory {
...
+    pub missing_criteria: Vec<CriteriaName>,
+}
+
+/// We recommend auditing the following package
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct JsonSuggestItem {
+    /// The name of the package
+    pub name: PackageName,
+    /// Any notable parents the package has (can be helpful in giving context to the user)
+    pub notable_parents: String,
```
mystor, line 631: I realize this is existing, but it's a bit odd that we encode this as a string.

src/format.rs
```diff
@@ -532,3 +533,116 @@ pub struct CommandHistory {
...
+    pub vetted_partially: Vec<JsonPackage>,
+    /// These packages are exempted
+    pub vetted_with_exemptions: Vec<JsonPackage>,
+}
+
+/// Failure! The violations and audits/exemptions are contradictory!
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct JsonReportFailForViolationConflict {
+    /// These packages have the following conflicts
+    pub violations: SortedMap<PackageAndVersion, Vec<ViolationConflict>>,
```
mystor, line 591: This is existing, but the fact we expose the internal `ViolationConflict` struct isn't the best API here.

Could we perhaps put a comment on some of these suboptimal API parts noting that we probably want to change this part of the json output? Now that we're actually specifying it, and building more complex tools on it, I'm more afraid about people depending on the current janky format and making it tricky for us to improve it.

mystor (approved):
This is a clear improvement over what we have, I think it does a bulk of what I wanted from #312 (other than having documentation in the book), but also does bring some highlights onto the weird parts of the output format.

It the bot workflow is using any of the specific parts of the json output which are a bit weird and I commented on, it might be nice to refine those outputs into something less weird sooner rather than later so that changing it isn't quite as painful in the future, but that doesn't need to block merging this.

Gankra:
I added FIXMEs for the places you pointed out.

I agree we should do a second pass on the schema before shipping an MVP of the bot/webapp. I don't expect the MVP to ship soon (lots of back and forth discussion and iteration on the design is inevitable), so I think landing this as-is to streamline future work on the schema is a good idea.

(I have some vague ideas around auto-generating json schemas and stuff so maybe a new library/tool is in the cards here.)

PR #351, opened 2022-10-31
Add --output-format=json-full

Gankra:
This is based on top of #350 and adds the remaining pieces a bot needs for #330.

The main piece is that we add `--output-format=json-full` which includes extra `context` (notably criteria definitions) and forces suggest output on even when locked.

An early version of this also stuffed all of `cargo metadata` in `context` and a `--with-metadata` flag so that cargo-vet could operate completely headless. Then automation could just zip up `supply-chain` and forget about the original repo. This is "necessary" because applying changes to the `supply-chain` "requires" cargo-vet to do it (as we want it to GC exemptions still).

However this was deemed more complex than just running `cargo vet certify` in the user's CI environment, so now the only context needed is the criteria (but I left in the supply-chain path because it's just generally useful even if I don't have an immediate usecase).

Unlike nika's approach, I don't add any "json input" for certify, and instead opt for CI scripts that do that looped invocation (certify already has flags for everything, including skipping prompts). I think nika's approach is totally reasonable though, and there's definitely an argument for doing it so that less people need to have that extra CI script complexity.


mystor:
@Gankra Is there any chance you'd be willing to rebase this change?

Gankra:
Rebase was hella messy because of all the snapshot churn. That said, it seems like with the changes you made arguing 'shallow' can be removed, it's possible that "confidence" no longer is a concept that needs to be tracked, so maybe almost nothing needs to be added now..?

mystor:
> Rebase was hella messy because of all the snapshot churn. That said, it seems like with the changes you made arguing 'shallow' can be removed, it's possible that "confidence" no longer is a concept that needs to be tracked, so maybe almost nothing needs to be added now..?

Yeah, I'm removing confidence completely in #384, as well as reworking the algorithm a bit, because we've removed dependency-criteria from audits/exemptions so we don't need to do the complex blaming part anymore (we can pre-compute requirements up-front with just the policy). (It'd be lovely if you're up to take a look over those patches and make sure I didn't mess something up, but no worries if you're busy!).

I wonder if we could get away with including the context unconditionally in normal --output=json output if we don't need to include the confidence boolean anymore? Main other thing I could think of being included in json-full would be the full search-result outputs, but that feels perhaps a bit too invasive (and I'm not sure tools would really benefit from it).

> (Sorry this took so long, coming back from being... Very Sick)

No worries! Glad you're feeling better ❤️

Gankra:
Yeah making `context` conditional was more important back when it included all of `cargo-metadata`, but now this can definitely all just be made "always" without any fuss.

PR #364, opened 2022-12-21
Add Embark and Bytecode Alliance to the registry

bholley:
We haven't implemented the code to fetch the registry from the tool, but it's still a useful place for discoverability.


mystor:
lgtm

Issue #360, opened 2022-12-08
FR: allow disambiguation of `audit-as-crates-io = true` crates not pulled through crates.io


gburgessiv:
The vast majority of third-party crates that we pull in are downloaded through `crates.io`, and `audit-as-crates-io` Does The Right Thing for much of the remainder. That said, the behavior specified on [first-party code docs](https://mozilla.github.io/cargo-vet/first-party-code.html) leaves the potential for audit gaps:

```
[dependencies]
# If `anyhow` at this revision were published onto crates.io, it would appear as 1.0.66.
# However, the sources may not equal 1.0.66 as published.
# anyhow = { git = "https://github.com/dtolnay/anyhow", rev = "60e88002dcf7315487cbb2c724c66d2297c7bf46" }
anyhow = "=1.0.66"
```

Currently, the audit for the `git`-based `anyhow` would apply to the published `anyhow`, and vice-versa. It'd be nice if there were a mechanism for `cargo-vet` to differentiate between these somehow. Especially since the documentation notes:

"""
It's important to note that any audits for such derived crates still correspond to the crates.io version. This is what inspect and certify will display, and this is what you should review before certifying, since others in the ecosystem may rely on your audits when using the original crate without your particular modifications.
"""

Looking around, maybe #170 is relevant here?

bholley:
I agree we should improve this.

When we designed the audit-as-crates-io feature, we were primarily thinking about `[patch]`/`[replace]`. The thinking was that any deviations from the published version would be first-party commits authored by the project owners, and thus safe by construction.

However, things get more fuzzy with git dependencies, which are sometimes just used as an out-of-tree variant of `[patch]`, but sometimes used to pull in pre-release versions of third-party crates. This creates some ambiguity around the rules of engagement, which we would ideally resolve by having cargo-vet keep track of things.

One way to characterize the situation is that cargo-vet operates at published-version granularity, but there is a finer level of intra-version granularity that can be expressed with git revs. We could potentially solve this by allowing audits to optionally correspond to git revs. Ideally we'd only allow this in the delta case, so that users are nudged to record the bulk of their audit findings in version space (which is more likely to be reusable by others in the ecosystem).

@mystor WDYT?

mystor:
So the idea here is that there'd be some syntax by which we could use a git revision instead of a version number in audits? Would the assumption continue to be that git revisions are `audit-as-crates-io = false` unless specified otherwise?

[...]

PR #371, opened 2023-01-13
Filter diffs to allow ignoring certain generated files

mystor:
This will be more important once we support git revisions, as we won't be able to generate these files for git revisions, leading to unnecessarily large diffs.

As we want to re-generate diffstats to use the new format, this also updates the diffstat version number, and takes the opportunity to change the format to a more flexible one. This unfortunately leads to large test output changes, including JSON output, making it a breaking change.


mystor:
This is the first part of the patch stack for #360. I'm putting it up separately as it's a substantial change which is independently useful, and will be easier to review separately.

Here's hoping that everything behaves correctly on Windows - I've tried to make sure that I don't depend on anything unix-specific, but I might have missed some oddity of how git-for-windows handles paths. I made sure to add a test which should hopefully reveal if something is going to break.

I think there may be a small risk around someone on a case-insensitive filesystem declaring a file like `caRgo.LocK`, as I didn't do a case-insensitive compare of the filenames, but I'm hoping it's not an issue in practice, as these files are created by Cargo.

PR #377, opened 2023-01-18
Book updates (Jan 2023)

bholley:
- Fix and clarify description of default build-criteria policy.
- Document new precise audit feature for audit-as-crates-io git dependencies.
- Remove documentation for dependency-criteria on audits and exemptions.


bholley:
@hpenne FYI

book/src/audit-entries.md
```diff
@@ -52,45 +52,3 @@ This field is optional, but encouraged for two reasons:
...
-you want, but occasionally you may wish to add or remove criteria for certain
-subtrees.
-
-For example, a dependency used to encrypt sensitive data might need review from
-cryptography experts:
-
-```
-[audit.mynetworkingcrate]
-version = '2.3.4'
-dependency_criteria = { hmac: ['safe-to-deploy', 'crypto-reviewed'] }
```
hpenne, line 72: Now that the detailed description of the format of "dependency_criteria" is removed from here, does not not need to be added to "config.md" instead, so that users know what format it should be?

hpenne, line 72: It seems like the format is wrong. I tried this syntax, and the toml parser insists that i replace the ":" with a "=":
`
dependency_criteria = { hmac = ['safe-to-deploy', 'crypto-reviewed'] }
`

hpenne (changes requested):

book/src/config.md
```diff
@@ -140,9 +140,7 @@ For top-level crates, defaults to `safe-to-run`.
 
 #### `dependency-criteria`
 
-Allows overriding the above values on a per-dependency basis. Similar in format
-to the [equivalent field](audit-entries.md#dependency-criteria) in audit
-entries.
+Allows overriding the above values on a per-dependency basis.
```
mystor, line 143: As @hpenne mentioned, it'd be nice to move some of the examples from the `audit-entries.md` file.

It's probably worth also updating the wording for the description of the policy table to reflect that policy fields are respected for both first-party and third-party/audit-as-crates-io crates (currently it claims that they're only respected for first-party crates). IIRC currently we only respect `criteria` and `dev-criteria`, but we will respect `dependency-criteria` for third-party/audit-as-crates-io crates.

book/src/first-party-code.md
```diff
@@ -22,10 +22,21 @@ and requires the user to specify the intended behavior. Specifically, if there
...
-will perform audit enforcement as if the crates.io version were being used.
+will perform audit enforcement.
 
-It's important to note that any audits for such derived crates still correspond
-to the crates.io version. This is what `inspect` and `certify` will display, and
-this is what you should review before certifying, since others in the ecosystem
-may rely on your audits when using the original crate without your particular
-modifications.
+When enabled for a git dependency, this enforcement is precise. It requires an
+audit for the base published version that exists on crates.io, and then one or
```
mystor, line 28: It currently doesn't actually require an audit for the "base published version" (as we discussed, the git dependency might have an unpublished version or a version which is not as related to the actual version) - it's just acting like a normal revision.

This wording perhaps should be updated to instead suggest that things "should" be audited as deltas from the base version?

mystor (changes requested):

book/src/audit-entries.md
bholley: Good point.

bholley: Good catch, fixed.

mystor (approved):

PR #384, opened 2023-01-20
Remove uncertainty and the blaming step from the resolver

mystor:
This patch does some fairly-major changes to the underlying resolver algorithm. Thanks to the removal of dependency-criteria from audits and exemptions, it's now possible to pre-compute the required criteria for every crate in the dependency graph ahead-of-time, and then validate each crate's criteria separately.

This removes the need for the blame phase, as we already know the responsible packages ahead of time, allowing these phases to be disentangled from one-another and simplified.

As there is no more uncertainty, the CriteriaFailureSet type is no longer necessary, and the output of a significant number of commands has been re-ordered to reflect that.

In addition, the algorithm page in the book has been updated to reflect the new implementation.


src/resolver.rs
```diff
@@ -1335,602 +1174,410 @@ pub fn resolve<'a>(
...
-    }
+        let normal_criteria = requirements[pkgidx].clone();
 
-    // If there were violations, report that
-    if !violations.is_empty() {
-        return (
-            results,
-            Conclusion::FailForViolationConflict(FailForViolationConflict { violations }),
-        );
+        // For each dependency,
```
afranchuk, line 1229: I expect there should be more to this comment. Maybe
```suggestion
        // For each dependency, elaborate the dependency criteria from the configured policy and add it to the dependency requirements.
```
Though the above section which does it first for the `dev_deps` doesn't have any such comment :shrug: I would be equally satisfied with this lingering half-comment being removed.

afranchuk (approved):
From my mostly-unfamiliar view, this looks good. I did ignore some of the logically-different deleted lines when reviewing, since it would be onerous to read and truly comprehend the prior model to compare with the new one.

src/resolver.rs
mystor: Lol, that sure looks like a partial comment I got distracted while writing :p - thanks for the catch.

mystor:
Yeah, the diff is certainly not the most readable, especially with the scale of the changes in parts.

Issue #394, opened 2023-02-02
Enable Base64 decoding for googlesource.com


djkoloski:
We host all of our code in gitiles, which doesn't provide a way to directly download individual files. You can however download files as base64 blobs by adding `?format=TEXT` to the URL. It would be great if we could specify this as an option in our import section, something like:

```toml
[imports.fuchsia]
url = "https://fuchsia.googlesource.com/supply-chain/audits.toml?format=TEXT"
decode = "base64"
```

Is this something that would be worth adding to cargo-vet?

bholley:
As it happens, I ran into precise this issue earlier today trying to import the [ChromiumOS audits](https://chromium.googlesource.com/chromiumos/third_party/rust_crates/+/main/cargo-vet/audits.toml) into Firefox. Reading through the [relevant issue](https://github.com/google/gitiles/issues/7), I don't expect this to be resolved in gitiles anytime soon.

Given that this is likely a pretty unique need, I'd be inclined to just special-case it rather than adding a general flag. We could, for example, detect when we're fetching an import from googlesource.com, and if so, try base64-decoding it before parsing. If base64-decode fails, we'd just fall through to the regular parser in case the server behavior ever changes.

That said, given that multiple Google teams are now deploying cargo-vet, I wonder if we should just jump straight to deploying `cargo-vet aggregate`. That would spare external consumers from needing separate imports for ChromiumOS and Fuchsia (along with any other potential future deployments). The automation for that could handle the base64 encoding internally, and put the result in a dedicated location that was served as normal plaintext.

@djkoloski @gburgessiv thoughts?

djkoloski:
> Reading through the https://github.com/google/gitiles/issues/7, I don't expect this to be resolved in gitiles anytime soon.

I read through that gitiles issue as well and agree that it looks... less than hopeful. I'll take a look and see if we can get any motion on it though, since it seems to be widely desirable. It's worth a shot at least.

> I'd be inclined to just special-case it rather than adding a general flag.

I think this could be a good solution in the immediate-term.

> I wonder if we should just jump straight to deploying `cargo-vet aggregate`.

I think we should do that. I'll follow up with some of our internal teams and see how we can approach resource allocation for it, since that's the cloudiest part. I anticipate that this could take some time though.

afranchuk:
[...]

PR #397, opened 2023-02-06
Add base64 decoding for gitiles.

afranchuk:
This is a workaround for #394. Tested with googlesource.com.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,6 +17,7 @@ exclude = [
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
+base64-stream = "1.2.7"
 cargo_metadata = "0.15.2"
 clap = { version = "3.2.6", features = ["derive"] }
 clap-cargo = "0.9.1"
--- a/src/network.rs
+++ b/src/network.rs
@@ -9,10 +9,12 @@
 
 use std::{
     ffi::{OsStr, OsString},
+    io::Write,
     path::{Path, PathBuf},
     time::Duration,
 };
 
+use base64_stream::FromBase64Writer;
 use reqwest::{Client, Url};
 use tokio::io::AsyncWriteExt;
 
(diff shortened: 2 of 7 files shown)


afranchuk:
If we don't want to add more exemptions, I can audit the added packages.

src/network.rs
```diff
@@ -29,6 +31,47 @@ static DEFAULT_TIMEOUT_SECS: u64 = 60;
...
+impl From<&Response> for PayloadEncoding {
+    fn from(response: &Response) -> Self {
+        // gitiles always encodes content in base64
+        if response.headers().contains_key("x-gitiles-object-type") {
+            Self::Base64
+        } else {
+            Self::Plaintext
+        }
+    }
+}
```
mystor, line 55: nit: as this is never really used with `into`, I'd probably just write this as a normal constructor like `fn for_response(response: &Response) -> Self {` rather than a `from` impl, as that lets you give more context.

Cargo.toml
```diff
@@ -17,6 +17,7 @@ exclude = [
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
+base64-stream = "1.2.7"
```
mystor, line 20: TBH I'm a little surprised that the `base64` crate (which we already depended on transitively through `reqwest`) doesn't support this, but it does seem like it only has a [decoding reader](https://docs.rs/base64/0.21.0/base64/read/index.html) and an [encoding writer](https://docs.rs/base64/0.21.0/base64/write/index.html), with no decoding writer or encoding reader.

Doing an review of `base64-stream` itself might be worthwhile as it looks like it contains some unsafe code but is fairly digestible, but we don't require every new dependency be audited yet for cargo-vet.

mystor (approved):
FWIW it would've probably been fine for perf if we read the entire response in before base64-decoding it, as it's not like we're using the data in a streaming way, but this is fine too :-)

src/network.rs
afranchuk: Sure. I usually add `From` impls if a type can be constructed from a single other type, but I don't mind adding an explicit method.

Cargo.toml
afranchuk: I was also surprised. I think the `base64` crate should expose structs which store the state needed for streamed encoding/decoding and allow explicit data IO so you don't necessarily need to have a `Read` or `Write`, too (which would make support for async stuff easier). Oh well.

I can review `base64-stream` (I've already read over a few files in it anyway).

PR #398, opened 2023-02-06
Wildcard audits

### LINKED ISSUE #353: RFC: Wildcard Audits
When an organization publishes its own code as a third-party crate and then consumes it back into the main project, cargo-vet allows [self-certification](https://mozilla.github.io/cargo-vet/curating-your-audit-set.html#self-certification). In practice this seems to be a [common](https://searchfox.org/mozilla-central/rev/2f47e3dacf0d773e9c7f363cecf10cfbea490679/supply-chain/audits.toml#513) [pattern](https://github.com/bytecodealliance/wasmtime/blob/3b6544dc66e704a185d7aaf8cc50bcad9d7c7f75/supply-chain/audits.toml#L606) for organizations today.

The current workflow for this is a little clunky:
* The author needs to touch audits.toml each time they go through a publishing cycle, which may require [explicit approval](https://groups.google.com/a/mozilla.org/g/governance/c/wMWBqkCnR34).
* If there's any delay between publishing the new version and pulling it into the main project, other consumers may find themselves importing it before the self-certification is available and end up [reauditing](https://phabricator.services.mozilla.com/D161748#change-fIGTpIsXGUmk).
* The certify UX isn't really optimized for the self-certification case.
* It feels like busywork, and pollutes the audit log with a bunch of repetitive and low-signal information.

We have thus far intentionally avoided a "just trust this crate" feature because there are lots of ways it could go wrong. However, I think I've come up with a set of guardrails that strike the right balance between ergonomics and security.

The feature would enable entries like this to `audits.toml`:

```
[[wildcards.wast]]
who = "Alex Crichton <alex@alexcrichton.com>"
criteria = "safe-to-deploy"
user-id = 1
start = "2019-10-1"
end = "2023-11-1"
notes = "The Bytecode Alliance is the author of this crate and ensures that all published versions are safe-to-deploy."
``` 

A few things to note:
* Like other content in `audits.toml`, these entries would be shareable by default, though we could provide a mechanism to ignore them when specifying the import. This is important to preserve the dynamic wherein small developers can piggy-back on the legwork of well-known organizations in figuring out what's safe.
* We restrict the wildcard to a single publisher, who should generally be the person on the "who" field. The idea is that this person is saying "any version of this package I publish in this date range will meet the provided criteria".
* We identify the publisher by crates.io user-id, rather than GitHub handle, because people can rename their GitHub handle and the old handle becomes available if they don't remember to squat it. We can maybe add some sugar for looking up your user ID in the `certify` workflow.
* We bound the wildcard by a publish date range, rather than a version range.
* We restrict the end date to 1 year into the future. This requires people to periodically renew their wildcards and avoids an indefinite blank check in cases where people drift away from the project.
* If a wildcard is needed in the vetting algorithm, it's reified into `imports.lock` to enable operation in `--locked` mode.
* API call results are cached in the cache directory, and only refetched if the registry index shows a new publish for the relevant crate (or the cache is cleared).

mystor:
Fixes #353 

This builds upon the changes in #399 to implement basic support for wildcard audits. These can be specified in your audits.toml file under the `wildcard-audits` key (rather than `audits`). A separate key is used to ensure that wildcard audits are placed above normal audits in audits.toml, to make them more discoverable.

Wildcard audits take a crates.io user-id and date range, and act as an audit for any crates published by the user with that user-id in that date range.

In order to avoid hitting the crates.io API too often, we only fetch publisher information about crates when needed, and maintain a local cache of publisher information in the cache directory.

While all information about publishers for each version is stored in the cache directory, only publisher information which is necessary will be stored in the imports.lock file for use in --locked mode. This is done in a similar manner to other imported audits since #399 by tracking which audit edges are used when running the resolver.

Wildcard audit end dates are allowed to be specified at most 6 months in the future.


mystor:
Leaving this as a draft until #399 is merged, as it contains all patches from that stack and will otherwise likely be difficult to review.

bholley:
Generally this looks nice! A few thoughts:
* I was initially surprised that the audit wasn't fully reified in imports.lock (e.g. no criteria). But then I realized that what we're really doing here is just caching the minimal crates.io information required for offline verification of the wildcard audit. That's actually a better model than what I'd been envisioning.
* Can we make the "when" field just YYYY-MM-DD rather than including the time, so that it aligns with the "start" and "end" fields?
* Was there a reason you opted for 6 months rather than 1 year? The latter feels a little more appropriate to me, and is also easy to reason about (if you want to certify as long as allowable, just take today's date and increment the year).
* What's our theory for how the user is supposed to discover the relevant crate id? I looked it up by inspecting the JSON at the API endpoint, but that's pretty obscure. It's true that we don't expect wildcard audits to be used as frequently, but it would be nice to make it as simple as possible. What about an interactive workflow like `cargo vet certify-wildcard CRATE USERNAME`? The command could do the id lookup, and also suggest some reasonable defaults (e.g. a start date of the earliest publish by that user, and and end date of one year in the future, assuming those are in fact the defaults we want to encourage).

mystor:
> Generally this looks nice! A few thoughts:
> 
>     * I was initially surprised that the audit wasn't fully reified in imports.lock (e.g. no criteria). But then I realized that what we're really doing here is just caching the minimal crates.io information required for offline verification of the wildcard audit. That's actually a better model than what I'd been envisioning.

Glad that this approach works well for you :-) - it's also more straightforward to implement than reifying audits, as it means we don't need to validate duplicated data against itself as often.

>     * Can we make the "when" field just YYYY-MM-DD rather than including the time, so that it aligns with the "start" and "end" fields?

Sure, I just put the full date in because it was available, but it's not a big deal to switch it over, and I suppose probably makes more sense as a thing to expose, as we don't really care about the time portion.

>     * Was there a reason you opted for 6 months rather than 1 year? The latter feels a little more appropriate to me, and is also easy to reason about (if you want to certify as long as allowable, just take today's date and increment the year).

I had a vague memory of you suggesting 6 months from the last time we talked, but I'm definitely up to go for 1 year instead - it's certainly simpler :-P

>     * What's our theory for how the user is supposed to discover the relevant user id? I looked it up by inspecting the JSON at the API endpoint, but that's pretty obscure. It's true that we don't expect wildcard audits to be used as frequently, but it would be nice to make it as simple as possible. What about an interactive workflow like `cargo vet certify-wildcard CRATE USERNAME`? The command could do the id lookup, and also suggest some reasonable defaults (e.g. a start date of the earliest publish by that user, and and end date of one year in the future, assuming those are in fact the defaults we want to encourage).

I don't have a great suggestion for how to discover user IDs right now, so we probably want to add a workflow like you mentioned. I think I'd prefer to add something like that in a follow-up though, as the patches are already complex enough as-is.

Adding a `certify CRATE --wildcard USERNAME` option or similar is probably a fairly straightforward option we could go with as well.

bholley:
Can you file a followup for the certify UX as well as updating the docs to describe this feature?

src/storage.rs
```diff
@@ -754,6 +877,29 @@ fn process_imported_audits(
...
+                for existing_audit in existing_audits {
+                    for new_audit in &mut *new_audits {
+                        // Ignore `who` and `notes` for comparison, as they
+                        // are not relevant semantically and might have been
+                        // updated uneventfully.
+                        if new_audit.is_fresh_import
+                            && new_audit.user_id == existing_audit.user_id
+                            && new_audit.start == existing_audit.start
+                            && new_audit.end == existing_audit.end
+                            && new_audit.criteria == existing_audit.criteria
```
afranchuk, line 895: I'd like to see this as a method on `WildcardEntry` so that if that is changed in the future, the potential to miss changing the corresponding equality is less.

src/storage.rs
```diff
@@ -986,6 +1157,141 @@ fn parse_imported_audit(valid_criteria: &[CriteriaName], value: toml::Value) ->
...
+        // to help decide whether or not to fetch new publisher information from
+        // crates.io, to reduce API activity.
+        let publishers = cache
+            .get_publishers(network, pkg_name, versions)
+            .map_err(Box::new)?;
+        relevant_publishers.push((pkg_name, publishers));
+    }
+
+    // NOTE: We make sure to process all imports before we look up user
+    // informaion in the cache, to ensure we're fetching consistent user
```
afranchuk, line 1256:
```suggestion
    // information in the cache, to ensure we're fetching consistent user
```

src/storage.rs
```diff
@@ -1144,16 +1462,25 @@ impl Cache {
             .and_then(|f| load_json(f).ok())
             .unwrap_or_default();
 
+        // Setup the command_history.
```
afranchuk, line 1465:
```suggestion
        // Setup the publisher_cache.
```

afranchuk (approved):
Looks pretty good to me, I've left a few comments.

src/format.rs
```diff
@@ -791,6 +854,79 @@ pub struct CommandHistory {
...
+pub struct PublisherCacheUser {
+    pub login: String,
+    pub name: Option<String>,
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct PublisherCacheVersion {
+    pub created_at: chrono::DateTime<chrono::Utc>,
+    pub num: semver::Version,
+    pub published_by: Option<u64>,
```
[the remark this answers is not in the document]
mystor: Sure

src/storage.rs
mystor: Sure, moving it to a `same_audit_as` method.

src/storage.rs
```diff
@@ -279,7 +294,9 @@ impl Store {
             imports_src: SourceFile::new_empty(IMPORTS_LOCK),
         };
 
-        store.validate(false)?;
+        let today = chrono::NaiveDate::from_ymd_opt(2023, 1, 1).unwrap();
```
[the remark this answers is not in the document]
mystor: Yeah, I think there's a solid chance we'll want this to be an argument in the future, I just didn't bother yet, because it'd be a bit of a hassle to change all of the call-sites.

src/format.rs
```diff
@@ -344,6 +356,33 @@ impl Serialize for Delta {
...
+/// These audits will be reified in the imports.lock file when unlocked.
+#[derive(Serialize, Deserialize, Clone, Debug)]
+pub struct WildcardEntry {
+    #[serde(default)]
+    #[serde(skip_serializing_if = "Vec::is_empty")]
+    #[serde(with = "serialization::string_or_vec")]
+    pub who: Vec<Spanned<String>>,
+    #[serde(with = "serialization::string_or_vec")]
+    pub criteria: Vec<Spanned<CriteriaName>>,
+    #[serde(rename = "user-id")]
```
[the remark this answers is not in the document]
mystor: No, I'm just mimicing the behaviour of other serializers which already exist. I think I'll leave it as-is for now, but we can perhaps clean it up in the future.

PR #405, opened 2023-02-14
Add an option to create wildcard audits with certify

### LINKED ISSUE #402: Add a flow to `certify` for creating wildcard audits
This flow should allow users to add new wildcard audits without needing to manually access the crates.io API in order to discover the unique user-ids for authors, instead enabling it to be looked up by a username. The command could probably look something like `cargo vet certify $PACKAGE --wildcard $USERNAME` with a flow to suggest and allow selecting start/end dates, as well as criteria, similar to the existing certify flow.

See https://github.com/mozilla/cargo-vet/pull/398#issuecomment-1423344354 for some initial thoughts on what this flow will look like.

mystor:
This adds a --wildcard flag to the certify subcommand which can be used to create wildcard audits. Start and end dates must be specified with --start-date and --end-date, though they have fallback values.

This is a first-draft of the UX, and we may improve it in the future. We expect creating wildcard audits to be less frequent than other types, so the experience is a bit less polished.

Fixes #402

### DIFF
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -245,12 +245,22 @@ pub enum CertifyError {
     #[error("'{0}' isn't one of your foreign packages")]
     #[diagnostic(help("use --force to ignore this error"))]
     NotAPackage(PackageName),
+    #[error("'{0}' has not published any relevant version of '{1}'")]
+    #[diagnostic(help("please specify a user who has published a version of '{1}'"))]
+    NotAPublisher(String, PackageName),
+    #[error("end date of {0} is too far in the future")]
+    #[diagnostic(help("wildcard audit end dates may be at most 1 year in the future"))]
+    BadWildcardEndDate(chrono::NaiveDate),
     #[error(transparent)]
     IoError(#[from] std::io::Error),
     #[error(transparent)]
     EditError(#[from] EditError),
     #[error(transparent)]
     UserInfoError(#[from] UserInfoError),
+    #[error(transparent)]
+    FetchAuditError(#[from] FetchAuditError),
+    #[error(transparent)]
+    CacheAcquire(#[from] CacheAcquireError),
 }
 
 ///////////////////////////////////////////////////////////
--- a/src/storage.rs
+++ b/src/storage.rs
@@ -20,12 +20,12 @@ use tracing::{error, info, log::warn, trace};
 
 use crate::{
     errors::{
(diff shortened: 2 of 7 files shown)


src/errors.rs
```diff
@@ -245,12 +245,21 @@ pub enum CertifyError {
     #[error("'{0}' isn't one of your foreign packages")]
     #[diagnostic(help("use --force to ignore this error"))]
     NotAPackage(PackageName),
+    #[error("'{0}' has not published any relevant version of '{1}'")]
+    #[diagnostic(help("please specify a user who has published a version of '{1}'"))]
+    NotAPublisher(String, PackageName),
+    #[error("end date of {0} is too far in the future")]
```
afranchuk, line 251: Maybe add a diagnostic help mentioning that it must be at most a year in the future?

src/storage.rs
```diff
@@ -798,6 +822,40 @@ impl Store {
...
+    /// Called to ensure that there is publisher information in the store's live
+    /// imports for the given crate. This is used when adding new wildcard
+    /// audits from `certify`.
+    pub fn ensure_publisher_versions(
+        &mut self,
+        cfg: &Config,
+        network: Option<&Network>,
+        package: PackageStr<'_>,
+    ) -> Result<&[CratesPublisher], CertifyError> {
+        info!("ensure_publisher_versions");
```
afranchuk, line 835: This would be more appropriate as `#[tracing::instrument]` on the function, if kept at all.

src/tests/certify.rs
```diff
@@ -147,3 +150,96 @@ fn mock_simple_certify_flow() {
...
+        None,
+        chrono::NaiveDate::from_ymd_opt(2023, 1, 1).unwrap(),
+    )
+    .expect("do_cmd_certify failed");
+
+    let audits = crate::serialization::to_formatted_toml(&store.audits).unwrap();
+
+    let result = format!("OUTPUT:\n{output}\nAUDITS:\n{audits}");
+
+    insta::assert_snapshot!("mock-wildcard-certify-flow", result);
```
afranchuk, line 244: The snapshot name here could be omitted.

afranchuk (approved):

src/storage.rs
mystor: Oops, thanks for finding my debugging logs which I forgot to remove :p

PR #409, opened 2023-02-15
Correctly handle importing audits for multiple versions

mystor:
When we changed the way that imports were handled in #399, the new minimal imports logic didn't handle multiple versions of crates with the same name properly, meaning that we'd only import the required audits for one of the versions rather than all of them.

This patch adds a test to ensure that doesn't happen again, and fixes the underlying issue.


afranchuk (approved):
Looks good.

PR #410, opened 2023-02-16
Avoid unrelated imports during certify and vet

### LINKED ISSUE #391: Consider pulling in imports via an explicit command
Right now, we re-check the import set every time somebody invokes a non-`--locked` cargo-vet command. If There's a new import that would allow us to remove an exemption, we automatically pull it in and drop the exemption.

Experience suggests that this is confusing for rank-and-file developers. They're usually in the process of auditing some dependency they added, and then some stuff for an unrelated crate appears in the same commit. Given all the ceremony of sign-off, it's a particularly bad moment for the tool to make mysterious changes without explanation.

I think we should switch to a model where cargo-vet simply prints a message after the "Vetting Succeeded" that says something like: `New audits are available for import — Please invoke cargo vet fetch-imports.` Or something like that.

mystor:
This is done by making whether imports or exemptions are preferred configurable on a per-package basis when running `regenerate_exemptions`.

This combines the existing iterative approach, which prefers imports, with an approach based on used edges when auditing, which prefers local audits. After the change, it is still possible for `vet` to cause imports to occur if they are necessary to vet successfully, however it will not cause unrelated exemptions to be removed.

Fixes #391

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -681,14 +681,6 @@ fn cmd_certify(
         today,
     )?;
 
-    // Minimize exemptions after adding the new `certify`. This will be used to
-    // potentially update imports, and remove now-unnecessary exemptions.
-    // Explicitly disallow new exemptions so that exemptions are only updated
-    // once we start passing vet.
-    match resolver::regenerate_exemptions(cfg, &mut store, false) {
-        Ok(()) | Err(RegenerateExemptionsError::ViolationConflict) => {}
-    }
-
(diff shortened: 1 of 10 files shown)


mystor:
Leaving as draft for now as this builds on #409

mystor:
I might be scrapping this code as part of implementing #411, so leaving this as a draft.

mystor:
Superceded by #412

PR #412, opened 2023-02-17
Avoid removing unnecessary imports and exemptions unless explicitly requested

### LINKED ISSUE #411: Avoid removing unnecessary imports and exemptions unless explicitly requested
After a discussion with wasmtime, it seems that the minimal imports approach from #399 of always using the absolute minimum possible set of imports may not be the best approach going forward. Instead, we want to try a hybrid approach which only imports the minimum set of new audits, but which also keeps any existing imports around until they're explicitly cleaned up by a command.

We could then emit a warning when running a command like `cargo vet` or `cargo vet suggest` directing the user to run a command to clean up unnecessary imports and exemptions, without performing unexpected changes when running normal vetting & certifying commands. One of the advantages of this is it allows workflows when interacting with external contributors where a core contributor can pre-vet/exempt a new crate or a new version of a crate for a contributor, and land the changes to the store before the contributor's patch lands without breaking the build.

This builds on top of the ideas from #391 by extending the logic to also apply to removing imports and exemptions which are no longer used, even by local commands. We will still want to have an option to perform exemption minimization for a single crate, potentially pulling in new imports, when `cargo vet certify` is run.

The current idea is to introduce a new command, `cargo vet prune`, which attempts to prune both exemptions and imports, with command line options to configure the behaviour. This may be able to replace the existing `cargo vet fetch-imports` (and potentially `cargo vet regenerate imports`) command(s).

mystor:
This patch changes the import and exemption minimization behaviour to be much more conservative than before, only performing full pruning of unnecessary audits and exemptions when running the new `cargo vet prune` subcommand (which replaces `cargo vet fetch-imports`), or when running `cargo vet regenerate {exemptions,imports}`.

When running other subcommands, the behaviour now avoids churn more.

* When running `cargo vet`, the command will only import new audits which are necessary to vet, but will not prune unused imported audits or unnecessary exemptions. This should reduce unintentional churn and avoid unrelated ride-along exemption removals with other changes.

* When running `cargo vet certify`, we will prune exemptions for only the newly certified crate, but not for any other crate. This takes advantage of the dependency-criteria removal which allows us to vet each crate fully independently.

In order to do this, the regenerate_exemptions backend has been fully rewritten. It now builds on top of the minimal imports backend, to avoid unnecessary churn, and uses flags passed to `search_for_path` to control audit traversal behaviour. Handling of exemption criteria expansion/minimization is done using extra ride-along CriteriaSet information on RequiredEdge nodes.

As these updates are no longer handled implicitly, we now emit a warning from `cargo vet` if a `cargo vet prune` command could allow unused exemptions or imports to be pruned. This is currently done somewhat sloppily by running both updates and comparing the outputs, but could be potentially refined in the future.

Fixes #411,#391

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -315,12 +315,12 @@ pub enum Commands {
     #[clap(disable_version_flag = true)]
     Fmt(FmtArgs),
 
-    /// Explicitly fetch the imports (foreign audit files)
+    /// Prune unnecessary imports and exemptions
     ///
-    /// `cargo vet check` will implicitly do this, so this mostly exists as "plumbing"
-    /// for building tools on top of vet.
+    /// This will fetch the updated state of imports, and attempt to remove any
+    /// now-unnecessary imports or exemptions from the supply-chain.
     #[clap(disable_version_flag = true)]
-    FetchImports(FetchImportsArgs),
+    Prune(PruneArgs),
 
     /// Fetch and merge audits from multiple sources into a single `audits.toml`
     /// file.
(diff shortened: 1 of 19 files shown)


afranchuk (approved):
Looks good, I like the flexibility of `update_store`.

PR #415, opened 2023-02-22
Ensure cargo-vet builds on rustc 1.61

mystor:
This is the current version used for macOS and windows toolchain builds in mozilla-central.


mystor:
Apparently us continuing to use 1.61 to build toolchains on macos and windows after https://bugzilla.mozilla.org/show_bug.cgi?id=1807765 was an oversight, and we should be using 1.65, which won't require the code changes from this patch to support.

mystor:
Closing for another PR to test on 1.65

PR #420, opened 2023-02-24
Update store after adding a wildcard audit

### LINKED ISSUE #419: Certifying a wildcard audit doesn't activate it until `vet check` is run again
@chutten pointed this out. For a non-wildcard certify, the audit is immediately applicable whether or not you run check again, because it's directly recorded in `audits.toml`. But for wildcard audits, it's not valid without the publisher details in `imports.lock`.

It seems like we kind of want an implicit `vet check` after wildcard certification, though I'm not sure if we'd run into trickiness in the case that vet still fails.

@mystor WDYT?

mystor:
This was accidentally missed when reworking how imports work, and should be handled for both wildcard and non-wildcard audits.

Fixes #419

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -1015,9 +1015,37 @@ fn do_cmd_certify(
         .iter()
         .map(|s| s.to_string().into())
         .collect();
-    let audit_kind = match kind {
-        CertifyKind::Full { version } => AuditKind::Full { version },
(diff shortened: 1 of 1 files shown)


afranchuk (approved):
Took me a second to realize the pertinent part was the removal of the `return`!

PR #421, opened 2023-02-27
Add some reference-level documentation for wildcard entries

mystor:


book/src/wildcard-audit-entries.md
```diff
@@ -0,0 +1,60 @@
+# Wildcard Audit Entries
+
+This section defines the semantics of the various keys that may be specified in
+wildcard audit table entries.
+
+Wildcard audits are an additional type of audit intended for organizations which
+which to publish self-audits for their crates. Using a wildcard audit, the
```
bholley, line 7: which which?

book/src/wildcard-audit-entries.md
```diff
@@ -0,0 +1,60 @@
...
+value is auto-populated from the git config.
+
+This field is optional, but encouraged for two reasons:
+* It makes it easier to attribute audits at a glance, particularly for
+  remotely-hosted audit files.
+* It emphasizes to the author that they are signing off on having performed the
+  audit.
+
+Note that the `who` user may be different than crates.io user specified by
+`user-id`.
```
bholley, line 55: I would say that it's technically allowed, but that the semantics here are that the person specified in `who` is making a promise about the behavior of the person specified in `user-id`, and so they either need to be the same person or have a close relationship (e.g., a team lead certifying a shared publishing account).

book/src/wildcard-audit-entries.md
```diff
@@ -0,0 +1,60 @@
+# Wildcard Audit Entries
+
+This section defines the semantics of the various keys that may be specified in
+wildcard audit table entries.
```
bholley, line 4: We should specify the table name somewhere, it's not really clear from the documentation.

book/src/wildcard-audit-entries.md
```diff
@@ -0,0 +1,60 @@
...
+
+A string identifying the auditor. When invoking `cargo vet certify`, the
+value is auto-populated from the git config.
+
+This field is optional, but encouraged for two reasons:
+* It makes it easier to attribute audits at a glance, particularly for
+  remotely-hosted audit files.
+* It emphasizes to the author that they are signing off on having performed the
+  audit.
+
```
bholley, line 53: This is copy-pasted from the equivalent entry for regular audits right? I'd just link to it rather than duplicating.

bholley (approved):
r=me with those fixes.

PR #426, opened 2023-03-03
Emit an error when trying to remap remote built-in criteria

mystor:
This was never properly implemented, and is being changed to emit a useful error, rather than producing confusing behaviour, such as accidentally creating an implies cycle between the safe-to-run and safe-to-deploy criteria.

This patch also removes documentation suggesting support for multiple criteria in criteria-map entries. This has never been implemented, and would not be trivial to implement with the current infrastructure, so has been removed. Some internal code has also been simplified to produce a better error in that situation.

### DIFF
--- a/book/src/config.md
+++ b/book/src/config.md
@@ -91,22 +91,25 @@ field is required.
 
 An inline table or array of inline tables specifying one or more mappings
 between the audit criteria of the imported and local sets. Each imported audit
-is matched against each mapping. If the imported audit certifies all of the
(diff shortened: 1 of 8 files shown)


afranchuk (approved):
Looks good. It'd be a somewhat complex feature to actually support a combination of `theirs`, and likely for little gain (i.e. the foreign project would likely have a single distinct criteria matching the semantics that somebody wants when making a mapping).

PR #429, opened 2023-03-07
Add chromeos to the registry

bholley:


bholley:
At some point we may switch over to an aggregated set for all of google, but this is worth making available for now.

mystor (approved):

Issue #104, opened 2022-05-31
Implement suggestions from the registry


bholley:
See https://mozilla.github.io/cargo-vet/performing-audits.html#suggestions-from-the-registry

bholley:
The registry is getting big enough that it's worth doing this now.

PR #434, opened 2023-03-10
Add support for registry suggestions

mystor:
This adds suggestions from the registry to the `cargo vet` and `cargo vet suggest` suggestion output. This is done by fetching the registry and the corresponding audits files when generating suggestions, and comparing the suggestion output from before and after to see if it would make an improvement.

Currently this adds no information to the JSON output to encode the registry suggestion information, though we may want to encode it somehow in the future.

Fixes https://github.com/mozilla/cargo-vet/issues/104


mystor:
Builds on top of #433, so leaving as a draft until that is merged.

mystor:
Also requesting a review from @bholley to look over the ux

bholley (approved):
I'm generally happy with the UX. I might come up with some proposed tweaks but this is plenty sufficient for a first stab.

src/resolver.rs
```diff
@@ -2161,6 +2166,131 @@ impl FailForViolationConflict {
...
+        if let Some(Reachable {
+            from_root,
+            from_target,
+        }) = reachable.as_mut()
+        {
+            // FIXME: this is horrible but I'm tired and this avoids false-positives
+            // and duplicates. This does the right thing in the common cases, by
+            // restricting ourselves to the reachable nodes that are common to all
+            // failures, so that we can suggest just one change that will fix
+            // everything.
```
afranchuk, line 2196: This isn't that bad IMO :)

src/resolver.rs
```diff
@@ -2161,6 +2166,131 @@ impl FailForViolationConflict {
...
+                    closest_below = Some(src);
+                }
+            } else if let Some(closest) = closest_above {
+                if src < closest {
+                    closest_above = Some(src);
+                }
+            } else {
+                closest_above = Some(src);
+            }
+        }
```
afranchuk, line 2255: I realize it'd be doubling the iterations, but I think this would be a lot more readable as:
```suggestion
        let closest_above = from_root.iter().filter(|src| src >= dest).min_by_key(|src| src - dest);
        let closest_below = from_root.iter().filter(|src| src < dest).min_by_key(|src| dest - src);
```

afranchuk (approved):

src/resolver.rs
mystor: Yeah, this isn't my comment, I just copied it from the original code.  It's fine.

src/resolver.rs
```diff
@@ -2007,7 +2009,7 @@ impl Suggest {
                 max1 = max1.max(console::measure_text_width(s1));
             }
 
-            for (s0, s1, s2) in strings {
+            for (s0, s1, s2, reg) in strings {
```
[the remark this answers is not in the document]
mystor: Fair enough, I think I'll tweak it a bit to simplify it slightly (as we're not doing alignment with the registry string so don't need it to be pre-computed) and then probably leave it though. I don't think doing that sort of refactoring is necessary quite yet.

src/resolver.rs
mystor: It actually occurs to me that `BTreeSet` is sorted already, so we can use the range syntax, and probably write it like:

```rust
        let closest_above = from_root.range(dest..).next();
        let closest_below = from_root.range(..dest).next_back();
```

Though it seems like this will require a turbofish to make it work with inference because of the reference - not the end of the world.

afranchuk: :facepalm: haha yeah that's much better. I didn't register they were `SortedSet`!

Issue #430, opened 2023-03-08
Enforce version consistency between tool and instance data


bholley:
When the data format of the files inside `supply-chain/` change in an incompatible way, we need to do a semantically-incompatible update. However, people will likely invoke cargo-vet on occasion against files that were generated by an earlier or later version of the tool, and we currently lack any principled way to specify behavior in those cases.

So the proposal is to record the major version of cargo-vet in `config.toml`. If the version in that file is higher than the version of the tool, the tool will abort and request an upgrade. If the version is lower, it will automatically bump it up and perform any necessary migration.

Issue #431, opened 2023-03-08
Eagerly map criteria into the local namespace


bholley:
Previously, we needed to represent foreign namespaces in the resolver for the purposes of dependency-criteria. Now that this feature has been removed, we can map them eagerly. This allows us to avoid importing the criteria descriptions and audit entries for criteria that are not directly mapped, reducing noise and also simplifying the implementation.

This will be a breaking change and thus require a major release.

PR #435, opened 2023-03-10
Track the version of cargo-vet used to create the supply-chain store

mystor:
This is important given changes like https://github.com/mozilla/cargo-vet/issues/431, which will change the meaning of imports.lock in such a way that earlier versions of cargo-vet may not understand parts of it correctly anymore. While we generally need to keep the audits.toml file somewhat forward-and-back compatible due to imports, this is
less of a concern for other files, so preventing downgrades could be valuable here.

This new approach allows implicit upgrades when running without --locked, but prevents downgrades or locked upgrades. cargo-vet versions with the same minor release are considered compatible, so only the major and minor components are considered.

Fixes https://github.com/mozilla/cargo-vet/issues/430

### DIFF
--- a/src/format.rs
+++ b/src/format.rs
@@ -1,6 +1,6 @@
 //! Details of the file formats used by cargo vet
 
-use crate::errors::VersionParseError;
+use crate::errors::{StoreVersionParseError, VersionParseError};
 use crate::resolver::{DiffRecommendation, ViolationConflict};
 use crate::serialization::spanned::Spanned;
 use crate::{flock::Filesystem, serialization};
@@ -420,6 +420,10 @@ impl WildcardEntry {
 /// config.toml
 #[derive(Debug, serde::Serialize, serde::Deserialize, Clone)]
 pub struct ConfigFile {
+    #[serde(rename = "cargo-vet")]
+    #[serde(default = "CargoVetConfig::missing")]
+    pub cargo_vet: CargoVetConfig,
+
     /// This top-level key specifies the default criteria that cargo vet certify will use
     /// when recording audits. If unspecified, this defaults to "safe-to-deploy".
     #[serde(rename = "default-criteria")]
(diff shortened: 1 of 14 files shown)


mystor:
Builds on top of https://github.com/mozilla/cargo-vet/pull/434, so leaving as a draft until that is merged.

src/format.rs
```diff
@@ -747,6 +751,109 @@ fn is_default_exemptions_suggest(val: &bool) -> bool {
...
+
+/// Cargo vet config metadata field for the store's config file.
+#[derive(Debug, Serialize, Deserialize, Clone)]
+pub struct CargoVetConfig {
+    pub version: StoreVersion,
+}
+
+impl CargoVetConfig {
+    /// Pretend that any store which was created without a version specified is
+    /// from version 0.3.
```
afranchuk, line 841: Shouldn't this be 0.4 (aren't we going to put this feature in 0.5)?

afranchuk (approved):
How will we handle migrations in the future if there are incompatibilities? I was expecting to see a strategy (boilerplate) in the code, though it's fine to omit it until we need it.

src/format.rs
mystor: I made it 0.3 because the current version of the main branch is 0.4, but yeah 0.4 makes more sense going forward.

mystor:
> How will we handle migrations in the future if there are incompatibilities? I was expecting to see a strategy (boilerplate) in the code, though it's fine to omit it until we need it.

I figured that pre-building a strategy for doing migrations before we know what we're going to need was a bit of pre-emptive optimization. So far we've never needed to e.g. break serialization and have managed to do migrations fairly painlessly, so I don't want to build an infrastructure for doing that before we need it. The main thing I was adding this for was for the interpretation of the machine-generated imports.lock file, which will be automatically updated the first time someone runs `cargo vet` with a new version because it'll be re-generated, so no explicit migration is necessary.

PR #438, opened 2023-03-14
Force fetching publisher information for non-dependency crates in certify

mystor:


bholley (approved):

bholley:
This fixes the issue that @kennykerr ran into. I'll put out a dot release.

bholley:
Hm, doesn't apply cleanly. We were just about to do an 0.5 release so I guess we can just bundle it into that.

Issue #458, opened 2023-04-19
RFC: Trusted Crates


bholley:
This RFC proposes a mechanism for projects to trust certain authors to publish specific crates, as well as a mechanism for these trust decisions to be surfaced to other projects as hints.

# Motivation

The purpose of cargo-vet is to achieve supply-chain integrity by ensuring that the contents of each dependency are signed off by a trusted source. The tool is audit-oriented because the crates in the ecosystem are very heterogeneous in origin: it's usually impractical to require that every dependency was _developed_ by a trusted source, so the next best thing is to ensure that everything has been _audited_ by a trusted source.

However, there are some cases where the developer is in fact trusted. In these cases, an audit isn't really needed. If _all_ your dependencies were authored by a trusted source, you wouldn't bother to deploy cargo-vet in the first place. Put differently: if I trust you to faithfully audit a third-party crate and certify that it's safe-to-deploy, I also trust you to make the same assessment about code you wrote. This is why cargo-vet permits organizations to [self-certify](https://mozilla.github.io/cargo-vet/curating-your-audit-set.html#self-certification) crates which they develop and publish.

Self-certification has proven very useful and provided a lot of crate coverage. Unfortunately, it doesn't easily scale: the author needs to submit a certification for their crate to the cargo-vet instance of a widely-trusted organization, which is a very manual process. In practice, the primary impetus for generating self-certifications occurs when a developer attempts to bring their own externally-developed crate back into a project they're working on. So for example, @alexcrichton has certified audits in the [wasmtime audit set](https://github.com/bytecodealliance/wasmtime/blob/main/supply-chain/audits.toml) for a number of external crates which he personally developed and which wasmtime uses, but it's a fraction of the full set of [314 crates](https://crates.io/users/alexcrichton) he's published on crates.io. Moreover, many/most authors of widely-used crates do not have an obvious place to record their audits — we've [occasionally](https://searchfox.org/mozilla-central/rev/31f5847a4494b3646edabbdd7ea39cb88509afe2/supply-chain/audits.toml#1528) [dabbled](https://github.com/bytecodealliance/wasmtime/blob/a486aa37ad187e4b5c5d34da303a39fa5cdd2745/supply-chain/audits.toml#L1757) with "guest audits", but that approach is neither elegant nor efficient.

[...]

PR #461, opened 2023-04-21
Trusted Crates Implementation 

mystor:
This is a first draft of an implementation of #458.

This is largely untested right now, and isn't ready to land, however it should be mostly feature-complete with the proposed design.


bholley (approved):
rs=me

src/resolver.rs
```diff
@@ -1512,6 +1557,34 @@ fn search_for_path(
...
 }
 
+/// Format a short list with commas and an "and" before the last item in a
+/// multi-item list. Also returns whether the list has multiple items.
+fn format_short_list(mut items: Vec<String>) -> String {
+    // To keep the display compact, sort by name length and truncate long lists.
+    // We first sort by name because rust defaults to a stable sort and this will
+    // have by-name as the tie breaker.
+    items.sort();
+    items.sort_by_key(|item| item.len());
```
afranchuk, line 1567: IMO this is easier to read as `items.sort_by_key(|item| (item.len(), item))`.

src/storage.rs
```diff
@@ -1313,6 +1335,26 @@ fn parse_imported_wildcard_audit(
     Some(audit)
 }
 
+/// Parse an unparsed wildcard audit entry, validating and returning it.
```
afranchuk, line 1338:
```suggestion
/// Parse an unparsed trust entry, validating and returning it.
```

PR #467, opened 2023-05-08
Support trusted entries in `cargo vet aggregate`

bholley:


mystor (approved):

mystor:
We should probably add some aggregate tests at some point, but this seems reasonable.

PR #469, opened 2023-05-08
Use crate metadata to consider whether audit-as-crates-io should be set for a crate.

### LINKED ISSUE #464: Avoid audit-as-crates-io errors when upstream publishes a new version
Over in Firefox, we recently encountered a bit of a [hiccup](https://bugzilla.mozilla.org/show_bug.cgi?id=1830488) with audit-as-crates-io.

The basic pathology was as follows:
1. Originally we were using [cssparser](https://crates.io/crates/cssparser) as an upstream crate.
2. We switched it to a git override to pull in some unreleased upstream changes, which included a version bump to 0.31.0
3. 0.31.0 was released upstream.
4. `cargo vet` started breaking with `Some non-crates.io-fetched packages match published crates.io versions`, and users are confused.

The source of confusion here is that the user implementing step (2) is different from the user encountering the error in step (4), and the problem can be precipitated at arbitrary times but another (potentially-unaffiliated) individual performing an out-of-band crates.io change in step (3). The ideal behavior is for the `audit-as-crates-io` entry to be required at step (2).

The underlying issue is that our criterion for "does this package need an audit-as-crates-io entry" is based on matching local versions with published versions, which is vulnerable to this exact kind of off-by-one failure mode.

I propose that we switch the "do you need an audit-as-crates-io entry" check from a version match to a description (or, if available, repository url) match. We already started using these fields in #441 to guess the correct audit-as-crates-io value during init/regenerate, but here I'm suggesting that we _assume_ that two same-name crates which share neither a description nor a repository URL across any published versions are unrelated to each other. This has the added benefit of reducing the number of audit-as-crates-io entries many projects will need [1].

We then need to deal with the fact that we'll now be permitting audits for versions of crates that don't exist on crates.io. For git dependencies, we now have an easy way to represent these with the `@git` syntax. For path dependencies, it's trickier — I think for now we can just require the user to rectifying the situation by requiring the user to either make the local version number match a published version (presumably the most-recently-published version), or set `audit-as-crates-io = false`.

[1] Aside from being a better similarity indicator, there's also a structural advantage: many local crates in Firefox have no description, whereas crates.io _requires_ a description for publication.

afranchuk:
This accesses the `crates.io` API, caching results in the publisher cache.

Closes #464.


src/storage.rs
```diff
@@ -1630,6 +1630,10 @@ impl Cache {
         }
 
         if cfg.mock_cache {
+            #[cfg(not(test))]
+            let publisher_cache = Default::default();
+            #[cfg(test)]
+            let publisher_cache = crate::tests::mock_publisher_cache();
```
mystor, line 1636: FWIW you could make this whole block `#[cfg(test)]` I think - `mock_cache` is test-only.

src/storage.rs
```diff
@@ -2226,27 +2230,21 @@ impl Cache {
...
+    /// When this function returns `Ok`, the returned state is guaranteed to have an entry for
+    /// `name` in `publisher_cache.crates`.
+    async fn update_publisher_cache(
         &self,
         network: &Network,
         name: PackageStr<'_>,
-        versions: FastSet<&semver::Version>,
-    ) -> Result<Vec<PublisherCacheVersion>, GetPublishersError> {
+        versions: Option<FastSet<&'_ semver::Version>>,
+    ) -> Result<std::sync::MutexGuard<CacheState>, GetPublishersError> {
```
mystor, line 2247:
```suggestion
        versions: Option<FastSet<&semver::Version>>,
    ) -> Result<std::sync::MutexGuard<'_, CacheState>, GetPublishersError> {
```

I like to include the lifetime placeholder explicitly.

src/storage.rs
```diff
@@ -2256,67 +2254,74 @@ impl Cache {
...
                         if missing_versions
-                            .iter()
-                            .all(|version| exact_version(&index_crate, version).is_none())
+                            .map(|v| {
+                                v.iter()
+                                    .all(|version| exact_version(&index_crate, version).is_none())
+                            })
+                            .unwrap_or(true)
                         {
                             info!("using cached publisher info for {name} - missing versions appear unpublished");
```
mystor, line 2291: Might be nice to change this `info!` log in the case where no versions are specified.

mystor (approved):
I think this should do what we're planning to do right now. Hopefully the behaviour changes don't cause any issues downstream going forward.

src/main.rs
```diff
@@ -2489,97 +2478,113 @@ fn foreign_packages_strict<'a>(
...
+                    continue 'restart;
                 }
-                // Now that we've found a version match, we're done with this package
-                continue;
+
+                shouldnt_be_audit_as.push(PackageError {
+                    package: package.name.clone(),
+                    version: Some(package.vet_version()),
+                });
             }
```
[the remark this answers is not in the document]
afranchuk: Yeah I think that'd be fairly unlikely, though that is essentially what the old problem was (changing the `version` such that it's inconsistent with crates.io state).

PR #470, opened 2023-05-09
Add --store-path to CLI to specify supply-chain directory.

jfgoog:
Our use case for this is the following:

* We want to import and build 3rd party Rust binaries -- such as `cargo-vet` itself.
* We want to apply our own auditing criteria, not necessarily what is use by the upstream maintainers.
* We want to leave the imported code untouched -- simply mirror the upstream repo, and not fork or maintain patches.

Without the ability to override the path to the supply-chain directory, we would have to fork or patch Config.toml to set `[package.metadata.vet]`.

As a side note, the behavior of `cargo vet` with respect to the current working directory differs from that of `cargo vendor`. `cargo vendor --manifest-path blah/Cargo.toml` uses a vendor directory in the current working directory, while `cargo vet init --manifest-path blah/Cargo.toml` creates a supply-chain directory in the same directory as the manifest.

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -33,10 +33,10 @@ pub struct Cli {
     #[clap(help_heading = "GLOBAL OPTIONS", global = true)]
     pub manifest_path: Option<PathBuf>,
 
-    /// Path to the audits directory
-    #[clap(long, name = "SUPPLY_CHAIN_DIR", parse(from_os_str))]
+    /// Path to the supply-chain directory
+    #[clap(long, name = "STORE_PATH", parse(from_os_str))]
     #[clap(help_heading = "GLOBAL OPTIONS", global = true)]
-    pub config: Option<PathBuf>,
+    pub store_path: Option<PathBuf>,
 
     /// Don't use --all-features
     ///
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -33,6 +33,11 @@ pub struct Cli {
     #[clap(help_heading = "GLOBAL OPTIONS", global = true)]
     pub manifest_path: Option<PathBuf>,
 
+    /// Path to the audits directory
+    #[clap(long, name = "SUPPLY_CHAIN_DIR", parse(from_os_str))]
+    #[clap(help_heading = "GLOBAL OPTIONS", global = true)]
+    pub config: Option<PathBuf>,
+
     /// Don't use --all-features
     ///
     /// We default to passing --all-features to `cargo metadata`
--- a/src/main.rs
+++ b/src/main.rs
@@ -419,7 +419,7 @@ fn real_main() -> Result<(), miette::Report> {
         })
         .transpose()?;
 
-    let cli_metacfg = cli.config.as_ref().map(|path| {
+    let cli_metacfg = cli.store_path.as_ref().map(|path| {
         MetaConfigInstance {
             version: Some(1),
             store: Some(StoreInfo {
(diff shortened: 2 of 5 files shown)


src/cli.rs
```diff
@@ -33,6 +33,11 @@ pub struct Cli {
     #[clap(help_heading = "GLOBAL OPTIONS", global = true)]
     pub manifest_path: Option<PathBuf>,
 
+    /// Path to the audits directory
+    #[clap(long, name = "SUPPLY_CHAIN_DIR", parse(from_os_str))]
+    #[clap(help_heading = "GLOBAL OPTIONS", global = true)]
+    pub config: Option<PathBuf>,
+
```
mystor, line 40: I believe that in the past we've used the term "store" (e.g. in the `[package.metadata.vet]` section) or "supply-chain" (e.g. as the directory name) to refer to this directory, rather than "config". Could we consider making this flag be `--store-path` instead?

This could probably also be documented as the "Path to the supply-chain directory" or similar.

tests/snapshots/test_cli__long-help.snap
```diff
@@ -1,5 +1,6 @@
 ---
 source: tests/test-cli.rs
+assertion_line: 104
```
mystor, line 3: I'm a bit surprised these lines are showing up in the yaml - is this a change with newer versions of `cargo insta`? IIRC they should only be in `.snap.new` files.

src/main.rs
```diff
@@ -419,12 +419,24 @@ fn real_main() -> Result<(), miette::Report> {
         })
         .transpose()?;
 
+    let cli_metacfg = cli.config.as_ref().map(|path| {
+        MetaConfigInstance {
+            version: Some(1),
```
mystor, line 424: I think this can be `None`, in which case it'll fall back to the default config. That way if a different config is specified in `Cargo.toml`, it will be respected.

mystor (changes requested):

src/cli.rs
jfgoog: Done.

tests/snapshots/test_cli__long-help.snap
jfgoog: Fixed. I had just copied the .snap.new files because I didn't see the instruction to run `cargo insta review`

src/main.rs
jfgoog: Yes, I think that's already what happens (with the fix below to precedence of --store-path), unless I am misunderstanding your comment.

src/main.rs
```diff
@@ -419,12 +419,24 @@ fn real_main() -> Result<(), miette::Report> {
...
         return Err(miette!("Both a workspace and a package defined [metadata.vet]! We don't know what that means, if you do, let us know!"));
     }
 
     let mut metacfgs = vec![default_config];
+    if let Some(metacfg) = cli_metacfg {
+        metacfgs.push(metacfg);
+    }
     if let Some(metacfg) = workspace_metacfg {
         metacfgs.push(metacfg);
     }
```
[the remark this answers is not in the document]
jfgoog: Done.

src/main.rs
```diff
@@ -419,12 +419,24 @@ fn real_main() -> Result<(), miette::Report> {
         })
         .transpose()?;
 
+    let cli_metacfg = cli.config.as_ref().map(|path| {
+        MetaConfigInstance {
+            version: Some(1),
+            store: Some(StoreInfo {
+                path: Some(path.clone())
+            })
```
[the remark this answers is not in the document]
jfgoog: Yes, that is already the behavior with this change.

When I put a crate in a subdirectory (src/) and run `cargo vet init --manifest-path src/Cargo.toml --store-path supply-chain --verbose info`, the supply-chain directory is created in the CWD, not inside src, and I see the following:

```
 INFO Final Metadata Config:
 INFO   - version: 1
 INFO   - store.path: Filesystem {
    root: "supply-chain",
}
```

src/main.rs
mystor: My comment is that right now it's treated as an explicitly specified version, rather than one which falls back, and it might be more elegant for it to fall back instead of being explicit. Doesn't really matter right now though I suppose, as there's only one version.

mystor (approved):

PR #473, opened 2023-05-15
Add the `renew` subcommand and a warning for expiring wildcard audits.

### LINKED ISSUE #465: Introduce nagging mechanism to update wildcard audit expiration
Wildcard audits have an expiration date (which can be at most one year in the future) so that the supply-chain owners will need to periodically reevaluate their relationship with the publisher and ensure the individual still carries the organization's stamp of approval. The rough expectation is that wildcard audits from individuals who are no longer associated with the project should be allowed to expire, though "associated" is a fuzzy concept and therefore some discretion is inevitable.

There are now dozens of wildcard audits in the ecosystem, all of which expire roughly one year after the feature was introduced. When that time comes around, most of them would ideally be renewed by bumping the `end` date. However, there's not a general mechanism to prompt people to do so. This could result in annoyance wherein project A depends on an imported wildcard audit from project B, but the person in project B hasn't bumped the expiration yet (perhaps because they're no longer using the crate within the project).

My proposal is to add a nag — similar to the `prune` nag — for any local wildcard audits that are expiring within the next six weeks:

```
WARN: Your audit set contains wildcard audits for foo, bar, and baz which expire within the next six weeks.
WARN:   Consider running `cargo vet renew --expiring` or adding `suggest-renewal = false` in audits.toml.
```

We'd then introduce the `cargo vet renew` command, which takes either an explicit crate name or the `--expiring` flag which operates on all wildcard audits expiring within the next six weeks.

afranchuk:
Closes #465.


src/main.rs
```diff
@@ -1649,6 +1657,76 @@ fn cmd_regenerate_audit_as(
...
+    let mut store = Store::acquire_offline(cfg)?;
+    do_cmd_renew(out, &mut store, sub_args);
+    store.commit()?;
+    Ok(())
+}
+
+fn do_cmd_renew(out: &Arc<dyn Out>, store: &mut Store, sub_args: &RenewArgs) {
+    assert!(sub_args.expiring ^ sub_args.crate_name.is_some());
+
+    let today = <chrono::DateTime<chrono::Utc>>::from(SystemTime::now()).date_naive();
```
mystor, line 1671: The point of these `do_...` methods is generally to allow testing, which means that we shouldn't be pulling system state like the timestamp within `do_cmd_renew` - could you make this an argument so it can be a fixed date in tests?

src/main.rs
```diff
@@ -1649,6 +1657,76 @@ fn cmd_regenerate_audit_as(
...
+    };
+
+    if let Some(name) = &sub_args.crate_name {
+        match store.audits.wildcard_audits.get_mut(name) {
+            Some(audits) => {
+                audits.iter_mut().for_each(renew_entry);
+                writeln!(
+                    out,
+                    "Updated wildcard audits for {name} to expire on {new_end_date}"
+                );
```
mystor, line 1685: It might be good to log how many wildcard audits for the crate were updated, and which users they were for, so that someone running the command can have an idea of who exactly they're updating the expiry date of the audits for?

src/main.rs
```diff
@@ -1649,6 +1657,76 @@ fn cmd_regenerate_audit_as(
...
+
+fn do_cmd_renew(out: &Arc<dyn Out>, store: &mut Store, sub_args: &RenewArgs) {
+    assert!(sub_args.expiring ^ sub_args.crate_name.is_some());
+
+    let today = <chrono::DateTime<chrono::Utc>>::from(SystemTime::now()).date_naive();
+    let new_end_date = today + chrono::Months::new(12);
+
+    let renew_entry = |entry: &mut WildcardEntry| {
+        entry.end = new_end_date.into();
+    };
```
mystor, line 1676: If the entry has `suggest-renewal = false` we should perhaps not renew it? Say there was a previous manager for a crate X, and you had them publishing the crate until they leave the company, then someone new takes over, so you mark the former employee's audit entry as `suggest-renewal = false` so it doesn't get renewed - it's kinda inconvenient that the renewal is happening automatically for both audits in that case.

Perhaps the member should instead be `renew = false` or something like that, and it can prevent the renew command from operating on the entry at all?

src/main.rs
```diff
@@ -1649,6 +1657,76 @@ fn cmd_regenerate_audit_as(
...
+
+        let mut updated = Vec::new();
+        for (name, entries) in store.audits.wildcard_audits.iter_mut() {
+            let any = entries
+                .iter_mut()
+                .filter(|e| e.should_suggest_renewal(expire_date))
+                .any(|e| {
+                    renew_entry(e);
+                    true
+                });
```
mystor, line 1704: Why the `.any` here? I think this behaviour for short-circuiting and only updating a single entry is unintuitive, as it's also different behaviour than is used for the explicit crate case (where every wildcard entry will be renewed).

Could you make this an explicit loop instead rather than using a side-effectful `.any` condition?

src/main.rs
```diff
@@ -1649,6 +1657,76 @@ fn cmd_regenerate_audit_as(
...
+            if any {
+                updated.push(name);
+            }
+        }
+
+        if updated.is_empty() {
+            info!("no wildcard audits are expiring in the next {WILDCARD_AUDIT_EXPIRATION_STRING}");
```
mystor, line 1711:
```suggestion
            info!("no wildcard audits have expired or are expiring in the next {WILDCARD_AUDIT_EXPIRATION_STRING}");
```

src/main.rs
```diff
@@ -1649,6 +1657,76 @@ fn cmd_regenerate_audit_as(
...
+        write!(
+            out,
+            "Updated wildcard audits for the following crates to expire on {new_end_date}:"
+        );
+        let mut iter = updated.iter();
+        write!(out, " {}", iter.next().unwrap());
+        for name in iter {
+            write!(out, ", {name}");
+        }
+        writeln!(out);
```
mystor, line 1726: Was this intended to use the new list formatting code you added?

Perhaps it'd make sense to do one crate on each line, and also write out who the publisher(s) are which you extended the wildcard audit(s) for?

src/main.rs
```diff
@@ -1970,12 +2048,57 @@ fn cmd_check(
             // more clever comparisons.
             if store.config.exemptions != pruned_exemptions {
                 warn!("Your supply-chain has unnecessary exemptions which could be relaxed or pruned.");
-                warn!("Consider running `cargo vet prune` to prune unnecessary exemptions and imports.");
+                warn!("  Consider running `cargo vet prune` to prune unnecessary exemptions and imports.");
```
mystor, line 2051: Why the extra two spaces?

src/main.rs
```diff
@@ -1970,12 +2048,57 @@ fn cmd_check(
...
+                    warn!("Your audit set contains wildcard audits for {expiring} which expire within the next {WILDCARD_AUDIT_EXPIRATION_STRING}.");
+                }
+                if !expired.is_empty() {
+                    let expired = errors::FormatShortList::new(expired);
+                    warn!(
+                        "Your audit set contains wildcard audits for {expired} which have expired."
+                    );
+                }
+                warn!("  Consider running `cargo vet renew --expiring` or adding `suggest-renewal = false` to the wildcard entries in audits.toml.");
+            }
```
mystor, line 2099: It'd be nice to pull this out into a separate function so that we can have some unit tests just on this logic.

book/src/wildcard-audit-entries.md
```diff
@@ -27,7 +28,11 @@ that releases are always cut from a branch for which every change has been
 approved by a trusted individual who will enforce the criteria.
 
 Wildcard audits can be added with `cargo vet certify` using the `--wildcard`
-option.
+option. This sets the `end` date to one year in the future. Once added (whether
+manually or by `cargo vet certify --wildcard`), the `end` date can be updated to
+one year in the future using the `cargo vet renew CRATE` command. `cargo vet
+renew --expiring` can be used to automatically update _all_ audits which would
+expire in the next six weeks or have already expired.
```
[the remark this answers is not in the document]
afranchuk: I'll reword that sentence. Currently there is no flag to change the end date for `renew` though, it's always 1 year. That could be changed.

I agree, I think it may make sense to set `suggest-renewal = false` if an explicit end date is provided.

src/cli.rs
```diff
@@ -681,6 +689,26 @@ pub struct GcArgs {
     pub clean: bool,
 }
 
+lazy_static::lazy_static! {
+    static ref RENEW_EXPIRING_HELP: String = format!("Renew all wildcard audits which will have expired {WILDCARD_AUDIT_EXPIRATION_STRING} from now.");
+}
+
```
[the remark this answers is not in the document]
afranchuk: Agreed, though printing help output does plenty of runtime string formatting. I could just put the words inline, at the cost of any changes to that duration being error-prone.

src/cli.rs
```diff
@@ -681,6 +689,26 @@ pub struct GcArgs {
...
+        action,
+        conflicts_with("crate-name"),
+        help = Some(RENEW_EXPIRING_HELP.as_str())
+    )]
+    pub expiring: bool,
+
+    /// The name of a crate to renew.
+    #[clap(value_name("CRATE"), action, required_unless_present("expiring"))]
+    pub crate_name: Option<String>,
+}
```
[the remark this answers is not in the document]
afranchuk: @bholley didn't want that (see https://github.com/mozilla/cargo-vet/issues/465#issuecomment-1546000571).

src/errors.rs
```diff
@@ -1021,3 +1021,106 @@ pub enum LoadJsonError {
...
+/// If there are 2 or fewer items, they are always displayed (regardless of formatting width
+/// limit), and the first item is also always displayed.
+///
+/// The formatting width limit pertains only to the total length of the items in the list, not
+/// additional characters/text added when formatting.
+///
+/// The default width is 40 characters.
+pub struct FormatShortList<S> {
+    items: Vec<S>,
+}
```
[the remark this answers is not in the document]
afranchuk: I agree, I was hoping for some input on that. There wasn't anything that was necessarily related to string formatting yet (I considered adding such a module). `out.rs` isn't bad though.

src/main.rs
afranchuk: Will do. I had a similar realization (though the tests I added aren't impacted by different dates). It'd be better to make it a parameter.

src/errors.rs
```diff
@@ -1021,3 +1021,106 @@ pub enum LoadJsonError {
...
+    }
+}
+
+impl<S: AsRef<str>> Display for FormatShortList<S> {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        use std::fmt::Write;
+
+        let width = f.width().unwrap_or(40);
+
+        // Below the character count is obtained with `len`, so these must be ASCII.
```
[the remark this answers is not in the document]
afranchuk: I saw that. I didn't think it would matter as there's no escape sequences (and from the perspective of this implementation, it's just string formatting).

src/errors.rs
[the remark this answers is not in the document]
mystor: I'd be fine with adding a new module too.

src/cli.rs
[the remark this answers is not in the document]
mystor: Yeah, I suppose it'd technically be possible to use `concat!(...)` or something like that, but it's pretty gross.

src/main.rs
afranchuk: Sure. It was originally adapted from a `.for_each`, so it became awkward.

src/errors.rs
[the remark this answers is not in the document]
mystor: I think the main thing would be unicode characters. Something like é can be written either as a single code point, or as a combination of multiple unicode code points (`"e\u0301"` vs `"\u00e9"`). The first case would be counted incorrectly for example.

I think we should do the right thing here even if it ends up working out OK.

src/main.rs
mystor: Might be nice to also capture a diff of the store in the tests as well, which would show the dates.

src/main.rs
mystor: Fair enough - the indentation seems fine.

src/main.rs
afranchuk: Ah good catch, I had intended to change that to be one per line. Including the publisher(s) would be a good addition.

src/main.rs
```diff
@@ -1970,12 +2048,57 @@ fn cmd_check(
...
             } else if store.imports != pruned_imports {
                 warn!("Your supply-chain has unnecessary imports which could be pruned.");
-                warn!("Consider running `cargo vet prune` to prune unnecessary imports.");
+                warn!("  Consider running `cargo vet prune` to prune unnecessary imports.");
             }
         }
+
+        // Check for wildcard audits which will be expiring soon, and warn about them.
+        {
+            let today = <chrono::DateTime<chrono::Utc>>::from(SystemTime::now()).date_naive();
```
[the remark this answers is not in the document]
afranchuk: Maybe we should fetch one consistent `today` in the `Config`?

src/main.rs
```diff
@@ -1970,12 +2048,57 @@ fn cmd_check(
...
+                warn!("  Consider running `cargo vet prune` to prune unnecessary exemptions and imports.");
             } else if store.imports != pruned_imports {
                 warn!("Your supply-chain has unnecessary imports which could be pruned.");
-                warn!("Consider running `cargo vet prune` to prune unnecessary imports.");
+                warn!("  Consider running `cargo vet prune` to prune unnecessary imports.");
             }
         }
+
+        // Check for wildcard audits which will be expiring soon, and warn about them.
+        {
```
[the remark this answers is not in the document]
afranchuk: Because the description of `locked` only mentioned that no new imported audits would be fetched. If it has a broader use case than the description implies I will change it to only check when not `--locked`.

src/main.rs
```diff
@@ -1970,12 +2048,57 @@ fn cmd_check(
...
+                if !expiring_soon.is_empty() {
+                    let expiring = errors::FormatShortList::new(expiring_soon);
+                    warn!("Your audit set contains wildcard audits for {expiring} which expire within the next {WILDCARD_AUDIT_EXPIRATION_STRING}.");
+                }
+                if !expired.is_empty() {
+                    let expired = errors::FormatShortList::new(expired);
+                    warn!(
+                        "Your audit set contains wildcard audits for {expired} which have expired."
+                    );
+                }
```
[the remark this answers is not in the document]
afranchuk: The thought crossed my mind too, happy to change it.

src/tests/renew.rs
```diff
@@ -0,0 +1,201 @@
...
+struct ExpireTest {
+    today: chrono::NaiveDate,
+    start: chrono::NaiveDate,
+    end: chrono::NaiveDate,
+}
+
+impl ExpireTest {
+    pub fn new(future: chrono::Duration) -> Self {
+        let today =
+            <chrono::DateTime<chrono::Utc>>::from(std::time::SystemTime::now()).date_naive();
```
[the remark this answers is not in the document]
afranchuk: Will do. As currently written this doesn't affect the tests, however making the tests pure is much better.

src/cli.rs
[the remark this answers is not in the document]
afranchuk: I'm going to change this to a doc string, because our other docs also have it non-programmatic anyway. It'd be nice if `clap` supported a `std::fmt::Arguments` for the help so you don't pay the price just to create a string that generally won't be used often.

src/errors.rs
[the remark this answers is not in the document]
afranchuk: Ugh, I wasn't thinking about unicode modifiers. I'll use `measure_text_width` then since it uses `unicode-width`. And then we can add terminal colors safely :)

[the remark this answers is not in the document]
afranchuk: Though then again I don't think most of the data going through this (like crate names) can have unicode, but better safe than sorry.

src/main.rs
afranchuk: I thought I had handled this case, and it turns it was partially handled. It would only ignore the `suggest-renewal` in the case where a crate name was supplied. But still, best to put a check in the function.

src/main.rs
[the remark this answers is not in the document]
afranchuk: Nope, it was easy!

afranchuk:
Still a few more things to do:
* add some store diff tests
* add some tests specifically for the recommendation logic (and refactor as necessary)
* should the suggestions happen with `--locked`?
* should `renew` be set to `false` if an explicit end date is provided?

mystor:
>     * [ ]  should the suggestions happen with `--locked`?

I'm vaguely inclined to not show them with `--locked`, but I suppose it doesn't matter much either way.

>     * [ ]  should `renew` be set to `false` if an explicit end date is provided?

I think so.

>     * [ ]  if a user runs `renew <crate>`, should this update end dates which are greater than the renewed date (i.e. decreasing the end date)?

This should never happen, as it is not allowed for a store to have a wildcard audit with an end date more than 1 year in the future - that is the maximum allowed value, and store validation will fail if it is exceeded.

src/main.rs
```diff
@@ -1649,6 +1650,66 @@ fn cmd_regenerate_audit_as(
...
+            return;
+        }
+    }
+
+    renewing.renew(new_end_date);
+
+    write!(
+            out,
+            "Updated wildcard audits for the following crates and user ids to expire on {new_end_date}:"
+        );
```
mystor, line 1698: This formatting is weird.

src/main.rs
```diff
@@ -1649,6 +1650,66 @@ fn cmd_regenerate_audit_as(
...
+    for (name, entries) in renewing.crates {
+        writeln!(
+            out,
+            "  {}: {:80}",
+            name,
+            // FormatShortList ends up sorting the strings by length and then
+            // lexicographically, which will serendipitously be the same as a numeric sort.
+            string_format::FormatShortList::new(
+                entries.iter().map(|a| a.0.user_id.to_string()).collect()
+            )
```
mystor, line 1708: It's a bit weird that we're showing the user_id number rather than the username or something like that, especially because the user id won't mean anything to most folks.

src/main.rs
```diff
@@ -1970,18 +2031,118 @@ fn cmd_check(
...
+    /// Whether there are no wildcard audits expiring or expired.
+    pub fn is_empty(&self) -> bool {
+        self.crates.is_empty()
+    }
+
+    /// Get the crate names for which wildcard audits have expired.
+    pub fn expired_crates(&'a self) -> Vec<PackageStr<'a>> {
+        self.crates
+            .iter()
+            .filter_map(|(name, ids)| ids.iter().any(|e| e.1).then_some(*name))
```
mystor, line 2122:
```suggestion
            .filter_map(|(name, ids)| ids.iter().any(|(_, expired)| expired).then_some(*name))
```

src/main.rs
```diff
@@ -1970,18 +2031,118 @@ fn cmd_check(
...
+            .collect()
+    }
+
+    /// Get the crate names for which wildcard audits will expire soon.
+    pub fn expiring_crates(&'a self) -> Vec<PackageStr<'a>> {
+        self.crates
+            .iter()
+            .filter_map(|(name, ids)| ids.iter().any(|e| !e.1).then_some(*name))
```
mystor, line 2130:
```suggestion
            .filter_map(|(name, ids)| ids.iter().any(|(_, expired)| !expired).then_some(*name))
```

mystor (approved):

src/main.rs
afranchuk: Agreed. It looks like for some reason cargo fmt doesn't want to reformat it. I know that can sometimes happen in macros but I'd expect `write!` to work.

mystor (approved):

PR #476, opened 2023-05-18
Rework the caching in crates.io metadata retrieval.

### LINKED ISSUE #472: Audit-as-crates-io detection needs some work
There are a few problems in the wake of #469 that I think we shouldn't ship:
* On big projects like Firefox and wasmtime, there's about a 1-minute spinnerless hang while cargo-vet hits the rate-limited crates.io API for all the first-party crates to see if there's a name collision. These results are cached, but only for 24 hours.
* The current flow leads users to set audit-as-crates-io=true for path crates with non-published versions, which is both semantically unsound and leads to problems elsewhere in the tool.

I think what we want is:
* Require public versions for path dependencies with audit-as-crates-io=true
* * Default to audit-as-crates-io=false when there's not a public version (at least during `init`, so that it always succeeds).
* Faster ability to quickly check whether there's a name collision by first hitting the non-rate-limited CDN, and only doing API lookups for crates that get a hit on the index.

afranchuk:
* Use index.crates.io rather than the local index.
* Add a spinner when checking audit-as-crates-io (which will be where the cache is typically updated).
* Rename the publisher cache to a more general crates-io-cache.

Closes #472.


src/storage.rs
```diff
@@ -2390,23 +2276,359 @@ impl Cache {
...
+        let url = Url::parse(&url).expect("invalid crate name");
+
+        let response = self.try_download(network, url).await?;
+
+        #[derive(Deserialize, Debug, Clone)]
+        struct IndexEntry {
+            vers: semver::Version,
+        }
+
+        let result = load_json_lines::<IndexEntry>(&response[..])?;
```
mystor, line 2584: You can get the full crates.io index values by using `crates_index::Crate::from_slice(...)` (https://docs.rs/crates-index/latest/crates_index/struct.Crate.html#method.from_slice), fwiw. So long as we're keeping that dependency around, there's not much point rolling our own deserialization.

src/storage.rs
```diff
@@ -2390,23 +2276,359 @@ impl Cache {
...
+        let entry = guard
+            .crates_cache
+            .crates
+            .entry(name.to_owned())
+            .or_insert_with(|| CratesCacheEntry {
+                last_fetched: self.now,
+                versions: Default::default(),
+                metadata: None,
+            });
+        entry.last_fetched = self.now;
```
mystor, line 2599: Given you'll be overriding `last_fetched` anyway, this could probably be `or_default()` I would imagine?

mystor (changes requested):
As I mentioned on Matrix, I'm concerned about our integration tests, and the fact we're checking the cache into tree. I think we'll probably be getting some updates to the cache frequently due to things expiring & being re-fetched, which means that there'll be a lot of churn on that file which we didn't have before.

We probably need either a way to prevent the cache from being updated when running these unit tests, or some way to avoid expiries on the cache (which seems somewhat impractical)

src/storage.rs
```diff
@@ -2390,23 +2276,359 @@ impl Cache {
...
+        network: &Network,
+    ) -> Result<std::sync::MutexGuard<'a, CacheState>, CrateInfoError> {
+        // Crate names can only be a subset of ascii (valid rust identifier characters and `-`), so
+        // using `len()` and indexing will result in valid counts/characters.
+        let mut url = String::from("https://index.crates.io/");
+        let name = self.crate_name;
+        use std::fmt::Write;
+        match name.len() {
+            1 => write!(url, "1/{}", name),
+            2 => write!(url, "2/{}", name),
```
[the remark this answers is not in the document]
afranchuk: Heh, hitting all the weird `write!` edge cases with fmt/clippy.

src/storage.rs
afranchuk: I didn't add a `Default` impl because I wasn't sure it made sense semantically to have `last_fetched` have a default. But the logic would all work fine if it happened to be left at that value (assuming `chrono` stuff has defaults that are the beginning of the representable time range).

afranchuk:
I've added a (hidden, but maybe it need not be?) `--current-time` flag which is now used in integration tests to have a fixed time. I tested the behavior by setting my system time a month in the future; there were no changes to the cache file when running tests.

src/main.rs
```diff
@@ -2695,90 +2708,127 @@ async fn check_audit_as_crates_io(
...
-                continue;
-            }
+    // We should only check the audit-as-crates-io entries if we have a network, because we
+    // shouldn't make recommendations based on potentially stale information.
+    if network.is_some() {
+        let progress = progress_bar(
+            "Fetching",
+            "crate metadata",
+            first_party_packages.len() as u64,
+        );
```
mystor, line 2733: A bit of me wonders if this should be more like `"Validating"`, `"audit-as-crates-io specifications"` so it's more clear why we're fetching metadata for these crates (especially because they're explicitly first-party crates). This name also works though.

src/storage.rs
```diff
@@ -1836,11 +1843,11 @@ impl Cache {
 
                 // Check if the resource is already available in our local cache.
                 let fetched_package_ = fetched_package.clone();
+                let now = filetime::FileTime::from_system_time(SystemTime::from(self.now));
                 let cached_file = tokio::task::spawn_blocking(move || {
                     File::open(&fetched_package_).map(|file| {
                         // Update the atime and mtime for this crate to ensure it isn't
                         // collected by the gc.
-                        let now = filetime::FileTime::now();
                         if let Err(err) =
```
mystor, line 1851: hmm, this is a bit weird because it does kinda mess with the mtimes of files in the local cache checkout, but that's probably fine as we aren't checking them in anyway? I don't think we run the gc on normal operations yet, but it might become more of a thing if/when we start doing that.

tests/test-cli.rs
```diff
@@ -135,6 +135,19 @@ fn test_markdown_help() {
...
+
+impl CommandExt for Command {
+    fn set_cache(&mut self) -> &mut Self {
+        self.arg("--cache-dir")
+            .arg("../cache")
+            .arg("--current-time")
+            .arg("2023-01-01 12:00:00Z")
+    }
+}
+
```
mystor, line 150: I wonder if we should make this also pass in `--manifest-path` and call it something like `test_project_args()`?

It looks like we have some `diff`-style calls which are probably running against the wrong repo right now because they don't specify the manifest-path, which is interesting, but should probably be fixed. (The `diff` command previously didn't care about the repo being acted upon, and these command lines are probably from when that was the case)

mystor (approved):

tests/test-cli.rs
afranchuk: I was curious whether those inconsistencies were intentional or not. It'd be good to make them all consistent.

src/storage.rs
afranchuk: This change gave me pause for a bit too, but I think for now it should be okay (and I _think_ it's more preferable than using some other time source?). It's simple enough to revert in the future.

mystor (approved):
Thanks for fixing up the versions errors.

PR #479, opened 2023-05-24
Simplify command for showing crate dependencies

bkrl:
Avoid useless use of `cat` by making `grep` read from `Cargo.lock` directly and replace `sort | uniq` with `sort -u`.


bholley (approved):
Thanks!

PR #480, opened 2023-05-26
Import additional registries

repi:
For auditing of `cargo-vet`'s dependencies itself, now including also Google's, ISRG's and Zcash registries. As suggested from `cargo-vet suggest`.

Increases amount of audited dependencies for 33 -> 46 which is a good step in the right direction.


bholley:
Thanks! This collides with a few changes I coincidentally pushed this morning, I'll fix it up.

bholley:
Fixed in another push.

Issue #481, opened 2023-05-26
`suggest` cmd can suggest same crate+version multiple times


repi:
For a few unaudited crates in our project running `cargo vet suggest` lists them multiple times even though they are the same version. Not a major issue though but just a bit odd so thought I should report it.

example:

```
    cargo vet inspect parking_lot 0.11.2                                                                      (used by bb8, sled, console-subscriber)                        (5584 lines)
    cargo vet inspect parking_lot 0.11.2                                                                      (used by cpal, eterm, pprof, tokio, and 49 others)             (5584 lines)
    cargo vet inspect bitflags 1.3.2                                                                          (used by ndk, nix, nix, nix, png, ron, and 45 others)          (2998 lines)
    cargo vet inspect bitflags 1.3.2                                                                          (used by minidump-writer, fxprof-processed-profile)            (2998 lines)
    cargo vet inspect memoffset 0.6.5                                                                         (used by nix, nix, ark-render)                                 (1007 lines)
    cargo vet inspect memoffset 0.6.5                                                                         (used by speedy)                                               (1007 lines)
    cargo vet inspect memoffset 0.6.5                                                                         (used by crossbeam-epoch, and 2 others)                        (1007 lines)
```

it also in some cases duplicates the same crate in the "used by", such has in `used by ndk, nix, nix, nix, png, ron, and 45 others` in the above, would have expected it to list `nix` just once there.

bholley:
@repi Thanks for the report, would be good to get that fixed.

Is it perhaps the case that (1) you have both `parking_lot 0.11.2 and `parking_lot X.Y.Z` in the tree, and (2) you have a delta audit for parking_lot `X.Y.Z -> 0.11.2`? That might plausibly cause this behavior (since each of the suggestions corresponds to a different node in your crate graph, but the suggested action happens to be the same).

If not, is there a way you could provide a minimized testcase?

repi:
we have 2 `parking_lot` crates used in this workspace and the lockfile, 0.11.2 and 0.12.1, the `memoffset` and `bitflags` crates we also have multiple versions of so that is probably the source of confusion here.

wrt to `parking_lot` we have a single audit imported for it:

```
[[audits.zcash.audits.parking_lot]]
who = "Jack Grigg <jack@z.cash>"
criteria = "safe-to-deploy"
delta = "0.11.2 -> 0.12.1"
[...]

PR #483, opened 2023-05-26
Filter out missing information in `get_publishers`

mystor:
This avoids a potential crash when the set of relevant versions does not force an update of crates.io API information, despite more version numbers being known.

This PR also contains a fix for #481, removing duplicated suggestions from the output.


bholley (approved):
rs=me

PR #492, opened 2023-06-09
Remove violations when pruning if no crates in the graph match the violation crate

### LINKED ISSUE #486: Violations should not appear in imports.lock if the crate is not used
Right now on m-c, cargo-vet wants to pull in a violation entry from embark for a crate (`epaint`) we don't use. I'm guessing that when we modified `imports.lock` to be more selective we missed violations somehow. We should fix that.

afranchuk:
If a crate does match, keep all violations (regardless of version).

Closes #486.

### DIFF
--- a/src/tests/import.rs
+++ b/src/tests/import.rs
@@ -220,8 +220,8 @@ fn existing_peer_skip_import() {
 
 #[test]
 fn existing_peer_remove_unused() {
-    // (Pass) When pruning, we'll remove unused audits when unlocked, even if
-    // our peer hasn't changed. These audits will be preserved when not pruning.
+    // (Pass) When pruning, we'll remove unused audits (including violations) when unlocked, even
+    // if our peer hasn't changed. These audits will be preserved when not pruning.
 
     let _enter = TEST_RUNTIME.enter();
     let mock = MockMetadata::simple();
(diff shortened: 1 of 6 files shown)


afranchuk:
Not sure what's up with the windows CI.

src/tests/import.rs
```diff
@@ -220,8 +220,8 @@ fn existing_peer_skip_import() {
 
 #[test]
 fn existing_peer_remove_unused() {
-    // (Pass) When pruning, we'll remove unused audits when unlocked, even if
-    // our peer hasn't changed. These audits will be preserved when not pruning.
+    // (Pass) When pruning, we'll remove unused audits (including violations) when unlocked, even
+    // if our peer hasn't changed. These audits will be preserved when not pruning.
```
mystor, line 224: Could we re-wrap this comment?

mystor (approved):

src/tests/import.rs
afranchuk: It is wrapped to 100 character lines. Want it shorter?

afranchuk: One of these days they'll stabilize comment wrapping in rustfmt...

PR #496, opened 2023-06-12
Support crates.io installs using the http crates.io registry

mystor:
Since Rust 1.70.0, the default protocol for the crates.io registry has been the sparse protocol. This protocol uses a different directory for its caches than the previous git-based protocol, and stores the index cache very differently on disk.

We already support using the sparse protocol for reading the index directly, however we were still relying on the non-sparse index on disk in order to locate cached source directories when computing diffs and generating suggest output.

This patch stops using crates-index to locate the crates.io index, as we don't need to handle more advanced features like custom indexes.  The simplified logic now checks for both the sparse and non-sparse crates.io indexes, and gracefully falls back if neither is found to downloading the crates ourselves.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -2699,11 +2699,6 @@ async fn check_audit_as_crates_io(
     network: Option<&Network>,
     cache: &mut Cache,
 ) -> Result<(), AuditAsErrors> {
-    // If we don't have a registry, we can't check audit-as-crates-io.
-    if !cache.has_registry() {
-        return Ok(());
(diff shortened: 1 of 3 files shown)


mystor:
This patch should also fix the windows CI failures which we observed in #492, as it is likely those errors occur because the CI machines no longer had a git registry.

afranchuk (approved):
I had a portion of this done before stepping away for the evening, but thankfully not all that much :)

PR #497, opened 2023-06-15
Audit unpublished audit-as-crates-io crates as a published version

### LINKED ISSUE #495: Better behavior for not-yet-published audit-as-crates-io crates
In the last release, we added better heuristics for detecting potential matches between first-party and published crates. Such potential matches require `audit-as-crates-io` entries. In wasmtime, there were ~50 [new entries](https://github.com/bytecodealliance/wasmtime/commit/5610cbf71068ba8d1154cff22b049561548c9ed9#diff-331b3f5fea27183313d109f512bc51e8a21ccd8584681b45c448bfa9af8d33eeR25).

These entries are new because wasmtime uses a branch-for-release-then-bump model, and so the local crates are always at least one version ahead of what's published on crates.io. This is a situation that we've struggled to handle well in cargo-vet. For git dependencies, we can require a precise audit chain all the way to the exact commit being imported. But for path dependencies, we currently just require `audit-as-crates-io` to be set to `false` if the local version is ahead of the published version.

The wasmtime case seems like motivation to do better here. In principle, I think the right way to support `audit-as-crates-io` in this situation is to require an audit for the most-recently-published version (that's also below the current version).

We could simply institute this behavior automatically (setting `audit-as-crates-io=true` causes the cargo-vet algorithm to just replace the explicit version of such a first-party crate with the most-recently-published version), though this would have the unfortunate effect of allowing a previously-passing `cargo vet` to start failing when a new version is published upstream.

I think we probably want to warn rather than error in that case, which means we need breadcrumbs of the previously-passing state. Not sure whether it's better to put those in the policy entry, in imports.lock, or if there's some more elegant situation.

@mystor WDYT?

mystor:
This is done by adding a new `unpublished` table to the imports.lock file, which will enumerate the mappings from local versions to published versions in use by the audit. These are treated like special delta audits when resolving, allowing the audit to pass.

Like other import types, these will be cleaned up when they are no longer necessary by `cargo vet prune`, though a new regenerate subcommand, `cargo vet regenerate unpublished`, will allow removing them even if they're still necessary.

In addition, if a smaller delta is available due to a new published version with an audit, `cargo vet prune` will also reduce that delta, to try to ensure that the audited version is as close as possible to the version in-tree.

When running `cargo vet suggest`, audits allowing the removal of outdated unpublished entries will be suggested, much like how audits are suggested to remove exemptions. Unpublished versions will never be suggested as an audit source or destination.

If the version becomes published, a warning will recommend running `cargo vet regenerate unpublished` to clean up the unpublished entry.

Fixes #495

### DIFF
--- a/src/format.rs
+++ b/src/format.rs
@@ -88,6 +88,12 @@ impl VetVersion {
             })
         }
     }
+
+    /// Check if this VetVersion exactly matches the given semver version with
+    /// no git revision metadata.
+    pub fn equals_semver(&self, semver: &semver::Version) -> bool {
+        self.git_rev.is_none() && &self.semver == semver
+    }
 }
 impl fmt::Display for VetVersion {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
@@ -906,6 +912,9 @@ impl Default for CargoVetConfig {
 pub struct ImportsFile {
     #[serde(default)]
     #[serde(skip_serializing_if = "SortedMap::is_empty")]
+    pub unpublished: SortedMap<PackageName, Vec<UnpublishedEntry>>,
+    #[serde(default)]
+    #[serde(skip_serializing_if = "SortedMap::is_empty")]
     pub publisher: SortedMap<PackageName, Vec<CratesPublisher>>,
     #[serde(default)]
     #[serde(skip_serializing_if = "SortedMap::is_empty")]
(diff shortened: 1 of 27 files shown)


src/format.rs
```diff
@@ -88,6 +88,12 @@ impl VetVersion {
             })
         }
     }
+
+    /// Check if this VetVersion exactly matches the given semver version with
+    /// no git revision metadata.
+    pub fn equals_semver(&self, semver: &semver::Version) -> bool {
+        self.git_rev.is_none() && &self.semver == semver
+    }
```
afranchuk, line 96: Could also add a `PartialEq<semver::Version>` impl.

src/resolver.rs
```diff
@@ -2419,17 +2476,20 @@ async fn suggest_delta(
...
-                            ..
-                        })
-                    )
+                // We always have sources for an empty crate.
+                let Some(ver) = ver else { return true; };
+                // We only have git sources for the package itself.
+                if ver.git_rev.is_some() {
+                    return ver == &package.version;
+                }
+                // We have sources if the version is has been published to crates.io.
```
afranchuk, line 2486:
```suggestion
                // We have sources if the version has been published to crates.io.
```

src/tests/mod.rs
```diff
@@ -1217,31 +1186,148 @@ fn diff_store_commits(old: &SortedMap<String, String>, new: &SortedMap<String, S
...
+    }
+}
+
+struct MockRegistryPackage {
+    versions: Vec<MockRegistryVersion>,
+    metadata: CratesAPICrateMetadata,
+}
+
+#[derive(Default)]
+struct MockRegistryBuilder {
```
afranchuk, line 1224: This is great, I was starting to think we needed something like this.

afranchuk (approved):

src/format.rs
mystor: Yeah, I felt like making it explicit made sense here because so that folks don't accidentally compare the semver component from e.g. a cargo_metadata package instead of calling `vet_version()`, but that'd probably also be fine.

PR #498, opened 2023-06-16
Fix URL when using crates.io index when package name contains uppercase

nbdd0121:


src/storage.rs
```diff
@@ -2557,6 +2557,8 @@ impl<'a> UpdateCratesCache<'a> {
             _ => write!(url, "{}/{}/{name}", &name[0..2], &name[2..4]),
         }
         .expect("writing to a String should not fail");
+        // Crate index always use lowercases, but crate name may contain uppercase characters.
+        url.make_ascii_lowercase();
```
mystor, line 2561: It looks like `cargo` itself [uses](https://github.com/rust-lang/cargo/blob/d93b018c20c72837b70b2f8dddc17f597528c77f/src/cargo/sources/registry/index.rs#L487) the `to_lowercase()` (unicode-handling) method rather than ASCII lowercase on the name string. That shouldn't generally matter though, as all crate names on crates.io are required to be ASCII right now (and it certainly doesn't look like [`make_dep_name`](https://github.com/rust-lang/cargo/blob/d93b018c20c72837b70b2f8dddc17f597528c77f/crates/cargo-util/src/registry.rs#L8-L20) would handle non-ascii strings). So this should be fine.

mystor (approved):

PR #500, opened 2023-06-19
Improve handling of trust --all

### LINKED ISSUE #499: cargo vet trust --all papercuts
Two papercuts I've discovered:

(1) It only seems to work on crates for which you have exemptions, but not for crates which are actively failing to vet. This seems like an oversight and not what we want.

(2) The "skipping X because it has multiple authors" can be kind of annoying. I think we should add a `--force` to override it.

mystor:
This patch makes a few changes to how trust --all is handled.

1. `trust --all` will now also trust crates which fail to vet, even if they don't have any exemptions.

2. A new flag, `--allow-multiple-publishers`, can be specified to also trust packages with multiple publishers.

3. `trust --all` will now more intelligently pick the default criteira based on your existing audit graph, providing better support for custom criteria.

Fixes #499

### DIFF
--- a/src/tests/certify.rs
+++ b/src/tests/certify.rs
@@ -674,3 +674,154 @@ fn mock_trust_flow_all() {
 
     insta::assert_snapshot!(result);
 }
+
+#[test]
+fn mock_trust_flow_all_allow_multiple() {
+    let mock = MockMetadata::simple();
+
+    let _enter = TEST_RUNTIME.enter();
+    let metadata = mock.metadata();
+
+    let (config, audits, imports) = files_inited(&metadata);
+
+    let output = BasicTestOutput::with_callbacks(|_| Ok("\n".to_owned()), |_| unimplemented!());
+
+    let cfg = mock_cfg_args(
+        &metadata,
+        [
+            "cargo",
+            "vet",
+            "trust",
+            "--all",
+            "testuser",
+            "--allow-multiple-publishers",
+        ],
+    );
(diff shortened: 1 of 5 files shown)


src/tests/certify.rs
```diff
@@ -674,3 +674,154 @@ fn mock_trust_flow_all() {
 
     insta::assert_snapshot!(result);
 }
+
+#[test]
+fn mock_trust_flow_all_allow_multiple() {
```
afranchuk, line 679: Maybe one more test which touches the code path which suggests the user add `--allow-multiple-publishers` would be good (it could be identical to this test but without the flag).

afranchuk (approved):

src/tests/certify.rs
mystor: The test above (`mock_test_flow_all`) does hit the codepath which adds the `--allow-multiple-publishers` code path. You don't see it in the captured output because the message is logged with `warn!` instead of to the `out`, so it doesn't appear in the test output unfortunately.

I considered looking into adding a new mechanism to register a `BasicTestOutput` as a handler for the tracing subscriber, but it's a bit tricky to do well (as the subscriber is shared between multiple test runs on multiple threads, happening concurrently - so connecting it back to the correct test is a bit sketchy). I don't know what the best way to have warnings like that be captured in the test output would be unfortunately. (probably would need to be TLS and bound to the current thread's warning logs only?)

PR #501, opened 2023-06-19
Suggest git audit failures in two stages

### LINKED ISSUE #378: When suggesting audits for a git revision, try suggesting audits for the base revision first
When a it revision fails to audit, and we make suggestions, we currently treat the git revision version like any other version, and suggest audits based on deltas.

Ideally, we'd like to encourage people to first audit the published "base version" of the crate being audited, and then produce a delta-audit from that "base version" to the git revision version separately, in order to produce more useful audits for sharing.

We could theoretically do this by running the resolver against the base version to see if it audits, checking if the base version can be found on crates.io, and if it can, and didn't audit, first suggesting the user audit the base version, and then separately suggesting they audit the git revision via. a delta.

This unfortunately adds a lot of complexity to the resolver and suggest algorithm, as we currently have no mechanism for running the resolver against a package not in the dependency graph. It may become simpler after #363, however, as we will be able to disentangle the core audit path-chasing logic and dependency graph requirement propagation logic if audit edges can't have custom dependency-criteria.

mystor:
This patch splits suggestions for git failures into two distinct stages,
one of which suggests that the closest published version be audited, and
then a suggestion of a delta-audit from that version to the git version.
In order to keep backlog numbers accurate, we need to suggest both the
delta and the base audit silmultaneously, which requires changes to how
suggestions are resolved.

Care was taken to ensure that these new suggestions should interact
correctly with other suggest features, such as registry suggestions and
trust hints. This is done by making these features only operate on the
"base" suggestion/published version when both are present.

Fixes #378

### DIFF
--- a/src/format.rs
+++ b/src/format.rs
@@ -381,6 +381,15 @@ impl Serialize for Delta {
     }
 }
 
+impl fmt::Display for Delta {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match &self.from {
+            Some(from) => writeln!(f, "{} -> {}", from, self.to),
+            None => self.to.fmt(f),
+        }
+    }
+}
+
 /// An entry specifying a wildcard audit for a specific crate based on crates.io
 /// publication time and user-id.
 ///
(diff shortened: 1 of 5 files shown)


mystor:
Leaving this as a draft for now, as it builds on the changes from #497, which haven't been merged yet.

afranchuk (approved):

PR #520, opened 2023-07-25
Add an importable flag to audits

mystor:
This adds a new option to audits, `importable`, which can be set to `false` to mark an audit such that it will be ignored when importing. This will be enabled by default for new audits containing git revisions, allowing those entries to be cleaned up when they are no longer necessary.

Currently this patch adds no mechanism to automatically collapse sequences of non-importable audits or similar, so cleanup will still need to be manual. This can be added in the future in a backwards-compatible way.

### DIFF
--- a/src/serialization.rs
+++ b/src/serialization.rs
@@ -303,6 +303,7 @@ pub mod audit {
         version: Option<VetVersion>,
         delta: Option<Delta>,
         violation: Option<VersionReq>,
+        importable: Option<bool>,
         notes: Option<String>,
         #[serde(rename = "aggregated-from")]
         #[serde(skip_serializing_if = "Vec::is_empty")]
@@ -334,6 +335,7 @@ pub mod audit {
                 notes: val.notes,
                 criteria: val.criteria,
                 kind: kind?,
+                importable: val.importable.unwrap_or(true),
                 aggregated_from: val.aggregated_from,
                 // By default, always read entries as non-fresh. The import code
                 // will set this flag to true for imported entries.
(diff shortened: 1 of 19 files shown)


src/serialization.rs
```diff
@@ -363,6 +365,7 @@ pub mod audit {
                 version,
                 delta,
                 violation,
+                importable: if val.importable { None } else { Some(false) },
```
afranchuk, line 368: I'm a bit concerned about this conversion because we lose whether an entry was explicitly set to `importable: true` in a round-trip serialization, which (in the future?) may have a different meaning than omission (e.g. we may have different rules about the default value when omitted). For instance, I was expecting `AuditEntry::importable` to be an `Option<bool>` and to see (at some point) something like
```
let entry: AuditEntry;
let is_importable = entry.importable.unwrap_or(!entry.kind.involves_git_revs());
```

But maybe you wanted to disallow ever importing git audits?

afranchuk (commented):
Looks good aside from the one case I brought up.

src/serialization.rs
mystor: The idea here is to be fully backwards compatible, meaning that no entry which was previously importable becomes non-importable after these changes. If the git entry was created before this patch is landed, it will not have `importable: false` specified on it, and so will be considered importable. Rather than making it so the default behaviour for git entries has been changed, this changes the code such that we specify `importable: false` for all new entries with git revisions.

afranchuk: I see, that sounds reasonable.

afranchuk (approved):

PR #521, opened 2023-07-27
Specify the number of chars to show with "git diff".

jfgoog:
On my system, "git diff" shows 8 hex digits by default, causing test_project_diff_output and test_project_diff_output_git to fail when I run "cargo test". I assume this is due to different versions of diff having different behaviors, but I have not verified.


bholley:
@mystor r?

PR #541, opened 2023-09-05
exclude inactive optional dependencies #537

TitanNano:
`cargo metadata` currently outputs all dependencies, even optional dependencies that are not active. By comparing `dep:*` features with dependencies that have been declared as optional, we can eliminate inactive optional dependencies.

For #537

### DIFF
--- a/tests/snapshots/test_cli__test-project-suggest-json.snap
+++ b/tests/snapshots/test_cli__test-project-suggest-json.snap
@@ -1449,7 +1449,7 @@ stdout:
       },
       {
         "name": "native-tls",
-        "notable_parents": "reqwest, hyper-tls, and tokio-native-tls",
+        "notable_parents": "hyper-tls and tokio-native-tls",
         "suggested_criteria": [
           "safe-to-deploy"
         ],
@@ -2025,7 +2025,7 @@ stdout:
       },
       {
         "name": "serde_json",
-        "notable_parents": "reqwest and test-project",
+        "notable_parents": "test-project",
         "suggested_criteria": [
           "safe-to-deploy"
         ],
@@ -2988,7 +2988,7 @@ stdout:
         },
         {
           "name": "native-tls",
-          "notable_parents": "reqwest, hyper-tls, and tokio-native-tls",
+          "notable_parents": "hyper-tls and tokio-native-tls",
           "suggested_criteria": [
             "safe-to-deploy"
           ],
@@ -3564,7 +3564,7 @@ stdout:
         },
         {
           "name": "serde_json",
-          "notable_parents": "reqwest and test-project",
+          "notable_parents": "test-project",
           "suggested_criteria": [
             "safe-to-deploy"
           ],
--- a/src/resolver.rs
+++ b/src/resolver.rs
@@ -188,9 +188,9 @@ pub struct PackageNode<'a> {
     /// The name of the package
     pub name: PackageStr<'a>,
     /// list of enabled optional dependencies
-    pub enabled_deps: Vec<&'a PackageId>,
+    pub enabled_deps: SortedSet<&'a PackageId>,
     /// list of optional dependency names
-    pub optional_deps: Vec<&'a PackageId>,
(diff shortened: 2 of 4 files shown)


bholley:
Thanks for the patch! @mystor WDYT?

tests/snapshots/test_cli__test-project-suggest-json.snap
```diff
@@ -1449,7 +1449,7 @@ stdout:
       },
       {
         "name": "native-tls",
-        "notable_parents": "reqwest, hyper-tls, and tokio-native-tls",
+        "notable_parents": "hyper-tls and tokio-native-tls",
```
mystor, line 1452: I think this change is incorrect. Looking at the `reqwest` crate, the default enabled features are [`["default-tls"]`](https://github.com/seanmonstar/reqwest/blob/798df70fb0a4cd786ecc8cf770b33f1d56528524/Cargo.toml#L29), which enables [`"native-tls-crate"`](https://github.com/seanmonstar/reqwest/blob/798df70fb0a4cd786ecc8cf770b33f1d56528524/Cargo.toml#L33), which in turn turns on the dependency on this crate (https://github.com/seanmonstar/reqwest/blob/798df70fb0a4cd786ecc8cf770b33f1d56528524/Cargo.toml#L112). In the test crate we directly depend on `reqwest` with default features from the test crate (https://github.com/mozilla/cargo-vet/blob/668c5bbf4dc14b9d9cef74458dc98181374ec5cc/tests/test-project/Cargo.toml#L10), so that crate and those features should definitely be enabled, and `reqwest` should be considered to be a parent crate of `native-tls`.

I have a suspicion that these dependency edges aren't showing up because the code is incorrectly assuming that crate names match the corresponding optional dependency feature names, which is not necessarily the case.

mystor (changes requested):
I don't think this change actually improves the situation here, and I'm not sure if there is actually an issue to start with.

When running `cargo metadata` from within `cargo-vet`, we intentionally enable all features within the root crate when auditing by default (we have a flag to disable that, if that's undesirable for your project: https://mozilla.github.io/cargo-vet/commands.html#--no-all-features).

`cargo metadata` already doesn't include dependencies which which are not required by features. For example, the `cargo_metadata` crate (which we use) has an [optional dependency on derive_builder](https://github.com/oli-obk/cargo_metadata/blob/184e836e3e197a07f5ea7eb30b5662163f0bc4ea/Cargo.toml#L15), however that feature is not enabled by cargo-vet. The non-required crate doesn't appear anywhere in our [Cargo.lock](https://github.com/mozilla/cargo-vet/blob/668c5bbf4dc14b9d9cef74458dc98181374ec5cc/Cargo.lock) and isn't required to be mentioned in our [supply-chain by cargo-vet](https://github.com/mozilla/cargo-vet/blob/668c5bbf4dc14b9d9cef74458dc98181374ec5cc/supply-chain/config.toml)

I think `cargo metadata` is already handling optional dependencies correctly, and only including them when needed. Do you have a concrete example where an optional dependency is incorrectly being included in the metadata? It would be useful to understand the problem we're trying to solve.

TitanNano:
> I think `cargo metadata` is already handling optional dependencies correctly, and only including them when needed. Do you have a concrete example where an optional dependency is incorrectly being included in the metadata? It would be useful to understand the problem we're trying to solve.

I have outlined one of our issues in a comment on the related issue: https://github.com/mozilla/cargo-vet/issues/537#issuecomment-1698863441

For example, sqlx is pulling in various drivers even though only one of them has been enabled via a feature.

There is also this cargo issue which indicates that this is indeed a problem, https://github.com/rust-lang/cargo/issues/10801. This all said, I can see that your example is correct, and I haven't yet identified what the difference to the sqlx example is.

tests/snapshots/test_cli__test-project-suggest-json.snap
TitanNano: This does indeed look like a false positive.

PR #566, opened 2023-10-16
Fix new clippy warnings/errors.

afranchuk:


afranchuk:
It looks like this has already been fixed.

PR #586, opened 2024-01-05
Perform import validation when aggregating remote audit files

mystor:
This copies the behaviour from normal imports, meaning that it handles things like version mismatches gracefully, and only emits a warning that those audits were ignored.

We may want to include an option to make this more aggressive, as aggregate is expected to be frequently run in automation so warnings may be invisible.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -15,14 +15,14 @@ use errors::{
     AggregateCriteriaImplies, AggregateError, AggregateErrors, AggregateImpliesMismatchError,
     AuditAsError, AuditAsErrors, CacheAcquireError, CertifyError, CratePolicyError,
     CratePolicyErrors, FetchAuditError, LoadTomlError, NeedsAuditAsErrors,
-    NeedsPolicyVersionErrors, PackageError, ShouldntBeAuditAsErrors, TomlParseError,
-    UnusedAuditAsErrors, UnusedPolicyVersionErrors, UserInfoError,
+    NeedsPolicyVersionErrors, PackageError, ShouldntBeAuditAsErrors, UnusedAuditAsErrors,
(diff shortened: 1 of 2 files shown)


src/storage.rs
```diff
@@ -1325,6 +1282,57 @@ pub(crate) fn foreign_audit_file_to_local(
...
 
+pub(crate) fn foreign_audit_source_to_local_warn(
+    name: &str,
+    audit_source: SourceFile,
+) -> Result<AuditsFile, LoadTomlError> {
+    // Attempt to parse each criteria and audit independently, to allow
+    // recovering from parsing or validation errors on a per-entry basis when
+    // importing audits. This reduces the risk of an upstream vendor adopting a
+    // new cargo-vet feature breaking projects still using an older version of
+    // cargo-vet.
```
afranchuk, line 1293: This might be appropriate as a doc comment on the function itself.

src/main.rs
```diff
@@ -2339,17 +2339,9 @@ fn cmd_aggregate(
...
-                    TomlParseError {
-                        source_code: audit_source,
-                        span: SourceOffset::from_location(&audit_string, line + 1, col + 1),
-                        error,
-                    }
-                })
-                .map_err(LoadTomlError::from)?;
+            let audit_source = SourceFile::new(&url_string, audit_string);
+            let audit_file =
+                storage::foreign_audit_source_to_local_warn(&url_string, audit_source)?;
```
afranchuk, line 2344: Maybe precede this call with a short comment describing why we're doing this here.

afranchuk (approved):

mystor:
The clippy lint failures will be fixed in #596

PR #588, opened 2024-01-17
Prefer local wildcard audits to remote (imported) audits.

### LINKED ISSUE #587: prefer local wildcard audits to remote regular audits
Over on [mozilla-central](https://phabricator.services.mozilla.com/D197955), @jrmuizel and @glandium identified an interesting issue where Jeff was trying to replace a regular audit with a wildcard audit. Rather than reifying the local wildcard audit, cargo-vet instead re-fetches the locally-deleted regular audit from the [aggregated mozilla-wide set](https://raw.githubusercontent.com/mozilla/supply-chain/main/audits.toml) and re-inserts it as an import.

I'm assuming this is because the precedence order is currently:
1. local regular audits
2. remote regular audits
3. local wildcard audits
4. remote wildcard audits

Ideally we'd swap 2 and 3.

@afranchuk Is my hypothesis correct?

afranchuk:
Closes #587.


src/tests/import.rs
```diff
@@ -1934,3 +1934,76 @@ fn import_criteria_map_aggregated_error() {
...
+            "third-party2".to_owned(),
+            vec![wildcard_audit(1, SAFE_TO_DEPLOY)],
+        )]
+        .into(),
+        trusted: SortedMap::new(),
+    };
+
+    imports
+        .audits
+        .insert(FOREIGN.to_owned(), foreign_audits.clone());
```
mystor, line 1973: To mode accurately model the situation which originally brought this up, we don't need any audits or their corresponding publishers pre-populated within `imports.lock`. We also don't need the foreign wildcard entry to be present.

src/tests/import.rs
```diff
@@ -1934,3 +1934,76 @@ fn import_criteria_map_aggregated_error() {
...
+
+    imports.publisher.insert(
+        "third-party2".to_owned(),
+        vec![publisher_entry_named(
+            ver(DEFAULT_VER),
+            1,
+            "user1",
+            "User One",
+        )],
+    );
```
mystor, line 1983: This publisher entry shouldn't be pre-populated within `imports.lock` - we should be fetching it from the network to model the situation better.

src/tests/import.rs
```diff
@@ -1934,3 +1934,76 @@ fn import_criteria_map_aggregated_error() {
...
+        audits: [(
+            "third-party2".to_owned(),
+            vec![full_audit(ver(DEFAULT_VER), SAFE_TO_DEPLOY)],
+        )]
+        .into(),
+        wildcard_audits: [(
+            "third-party2".to_owned(),
+            vec![wildcard_audit(1, SAFE_TO_DEPLOY)],
+        )]
+        .into(),
```
mystor, line 1967: I'm pretty sure this foreign wildcard audit is unnecessary, but I suppose it doesn't hurt to include it in the test.

src/tests/import.rs
```diff
@@ -1934,3 +1934,76 @@ fn import_criteria_map_aggregated_error() {
...
+        .package(
+            "third-party2",
+            &[reg_published_by(ver(DEFAULT_VER), Some(1), "2022-12-15")],
+        )
+        .serve(&mut network);
+    network.mock_serve_toml(FOREIGN_URL, &foreign_audits);
+
+    let store = Store::mock_online(&cfg, config, audits, imports, &network, true).unwrap();
+
+    let output = get_imports_file_changes_prune(&metadata, &store);
```
mystor, line 2007: Theoretically you shouldn't need to do a prune for the correct outcome to happen here.

src/resolver.rs
```diff
@@ -1446,6 +1447,7 @@ fn search_for_path(
     #[derive(Debug, Copy, Clone, Eq, PartialEq, Ord, PartialOrd)]
     enum CaveatLevel {
         None,
+        Import,
```
mystor, line 1450: I think this is ranked too high. A new import should never be ranked higher than something like a non-importable audit or preferred exemption/unpublished, as that means that unnecessary change to the imports.lock would happen without doing a prune.

mystor (changes requested):

src/tests/import.rs
afranchuk: Oh right, so I was testing in a situation where we did actually already have it there (import first, then add a wildcard audit). I can change this or make a separate test to test fresh imports.

src/tests/import.rs
afranchuk: The foreign wildcard entry was kind of to "cover all bases", though somewhat lazily.

src/tests/import.rs
afranchuk: We do need to prune for the correct outcome when the import is not fresh.

src/resolver.rs
afranchuk: A fresh import is marked `FreshImport`, this is for existing imports. They previously had `CaveatLevel::None`, so I thought this was appropriate.

afranchuk:
I've added a second test. Now one tests against existing imports and the other tests against a fresh import.

src/resolver.rs
```diff
@@ -1177,7 +1177,8 @@ impl<'a> AuditGraph<'a> {
                         audit_index,
                         publisher_index,
                     };
-                    let is_fresh_import = entry.is_fresh_import || publisher.is_fresh_import;
+                    let is_fresh_import = import_index.is_some()
+                        && (entry.is_fresh_import || publisher.is_fresh_import);
```
mystor, line 1181: (I meant to include a comment like this when I first submitted the review, but apparently it didn't save - apologies)

We shouldn't consider a local wildcard import which has a fresh import to be a non-fresh import. Something like that requires changes to `imports.lock` in order to use, so should be considered fresh.

Perhaps we could change `DeltaEdge::is_fresh_import` to instead be an enum like:

```rust
#[derive(Debug, Copy, Clone, Eq, PartialEq)]
enum DeltaEdgeFreshness {
  // All information requried for this delta edge is already stored within the store's supply-chain.
  // Using this edge will require no changes to imports.lock.
  Stale,
  // This edge originates from an already-imported or local wildcard audit,
  // however will require importing fresh publisher information into imports.lock to use.
  FreshPublisher,
  // This edge is fully fresh, and will require adding the audit entry to imports.lock to use.
  Fresh,
}

impl DeltaEdgeFreshness {
  fn is_fresh(&self) -> bool { self != DeltaEdgeFreshness::Stale }
}
```

For most of the logic, checks of `is_fresh_import` can be replaced with the equivalent of `is_fresh()` (or just inlining that check), but the caveat level match pattern would need to be changed, like:

```rust
                _ => match edge.is_fresh_import {
                   DeltaEdgeFreshness::Stale => CaveatLevel::None,
                   DeltaEdgeFreshness::FreshPublisher => CaveatLevel::FreshPublisher,
                   DeltaEdgeFreshness::Fresh => CaveatLevel::FreshImport,
               }
```

src/resolver.rs
mystor: I've added a bit more elaboration as to what I was going for in my comment above around this being too high of a ranking (tl;dr with the current implementation, this isn't just for existing imports, but also for local wildcard audits with fresh publisher entries, which definitely shouldn't be above these other cases).

I don't think we need a ranking for fully stale imports other than `None`, though I'm open to an argument for where that would be useful.

mystor (changes requested):

src/resolver.rs
afranchuk: This is the change I was most hesitant about, so I'm glad you had some opinions on it.

src/resolver.rs
afranchuk: If we don't rank fully stale imports other than `None`, a local wildcard audit won't be preferred over an existing import. So if you already have some imports set up but then add a local wildcard audit, the imports will remain and the local wildcard audit would be unused (and prunable). I think we want the local wildcard audit to be preferred.

src/resolver.rs
```diff
@@ -341,6 +341,40 @@ pub enum RequiredEntry {
...
+        }
+    }
+
+    pub fn fresh_publisher(is_fresh_publisher: bool) -> Self {
+        if is_fresh_publisher {
+            DeltaEdgeFreshness::FreshPublisher
+        } else {
+            DeltaEdgeFreshness::Stale
+        }
+    }
```
mystor, line 371: Perhaps this one should handle both sides, rather than just the publisher one?

Also not sure why these methods are public given the type they are on is not exposed outside of this module, so let's drop the `pub` annotations.

```suggestion
   fn fresh_publisher(is_fresh_audit: bool, is_fresh_publisher: bool) -> Self {
        if is_fresh_audit {
            DeltaEdgeFreshness::Fresh
        } else if is_fresh_publisher {
            DeltaEdgeFreshness::FreshPublisher
        } else {
            DeltaEdgeFreshness::Stale
        }
    }
```

src/resolver.rs
```diff
@@ -1203,18 +1242,19 @@ impl<'a> AuditGraph<'a> {
                     let to_ver = Some(&publisher.version);
                     let criteria = criteria_mapper.criteria_from_list(&entry.criteria);
                     let origin = DeltaEdgeOrigin::Trusted { publisher_index };
+                    let freshness = DeltaEdgeFreshness::fresh_publisher(publisher.is_fresh_import);
```
mystor, line 1245: I'm not sure we want to use `fresh_publisher` here. Technically it is somewhat "correct", but a trusted entry is less validated than a normal audit, so we might not want to be prioritizing them in this way?

I believe that if we put them at the same caveat level as other audits, it'll be picked below normal audits due to the tie-breaker ordering by `DeltaEdgeOrigin`, which is what we want:

https://github.com/mozilla/cargo-vet/blob/67e92b6d03f48802907b2c80119ed328289a114d/src/resolver.rs#L301-L302

https://github.com/mozilla/cargo-vet/blob/67e92b6d03f48802907b2c80119ed328289a114d/src/resolver.rs#L1486

```suggestion
                    let freshness = DeltaEdgeFreshness::fresh(publisher.is_fresh_import);
```

mystor (changes requested):

afranchuk:
@mystor does this look better now?

mystor (approved):

PR #590, opened 2024-02-01
Upgrade libz-sys in Cargo.lock

glandium:
In Firefox automation, we build using the Cargo.lock, and the latest libz-sys version contains a fix for clang 18 on mac (https://github.com/madler/zlib/commit/4bd9a71f3539b5ce47f0c67ab5e01f3196dc8ef9)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -818,9 +818,9 @@ dependencies = [
 
 [[package]]
 name = "libz-sys"
-version = "1.1.8"
+version = "1.1.15"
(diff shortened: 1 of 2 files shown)


bholley:
The cargo-vet run is failing. I'm ok if you want to just add it as an exemption.

glandium:
Err, of course, cargo-vet uses cargo-vet :)

bholley (approved):

PR #592, opened 2024-02-22
Additional tracing messages to diagnose missing audit criteria.

anforowicz:


anforowicz:
This PR helped me diagnose problems I've encountered when trying to import https://raw.githubusercontent.com/google/rust-crate-audits/main/manual-sources/google3-audits.toml into Chromium (see Chromium's [config.toml here](https://source.chromium.org/chromium/chromium/src/+/main:third_party/rust/chromium_crates_io/supply-chain/config.toml;l=47-62;drc=7c6a437d4824e968d0f7fd120f082ffbe5a22218)).

PTAL?

src/storage.rs
```diff
@@ -1233,21 +1234,33 @@ fn is_known_criteria(valid_criteria: &[CriteriaName], criteria_name: &CriteriaNa
...
-                ignored_criteria.push(criteria);
-                None
-            }
-        })
-        .collect();
-    let valid_criteria: Vec<CriteriaName> = criteria.keys().cloned().collect();
+    let valid_criteria: Vec<CriteriaName>;
+    let mut criteria: SortedMap<CriteriaName, CriteriaEntry>;
+    if foreign_audit_file.criteria.is_empty() {
+        warn!("No criteria definitions found when importing from `{name}`");
```
mystor, line 1244: I don't think this is a situation where we want to warn. We expect most organizations using cargo-vet to not need additional custom criteria beyond `safe-to-run` and `safe-to-deploy`, so there being no criteria definitions is an expected normal state.

src/storage.rs
```diff
@@ -1233,21 +1234,33 @@ fn is_known_criteria(valid_criteria: &[CriteriaName], criteria_name: &CriteriaNa
...
+                None => {
+                    ignored_criteria.push(criteria);
+                    None
+                }
+            })
+            .collect();
+        valid_criteria = criteria.keys().cloned().collect();
+        if valid_criteria.is_empty() {
+            warn!("All criteria definitions failed to parse when importing from `{name}`");
+        }
```
mystor, line 1262: The intention of `ForeignAuditFileToLocalREsult` is that it contains the relevant information (which criteria were ignored) in the returned data structure.

I would prefer to not add this extra logging in this code. If we want to add some form of error logging about this, I think it would be preferable to e.g. record how many criteria were present upstream in the `ForeignAuditFileToLocalResult` so the caller could log something like `{x} of {y} foreign custom criteria failed to parse when importing from {name}: {list}`.

If we want to add a `warn!` trace log within this function, I think I'd want it to be don in that style and based on `ignored_criteria` - perhaps something like:

```rust
if !ignored_criteria.is_empty() {
  warn!("{} of {} custom criteria failed to parse when importing from `{}`",
        ignored_criteria.len(), foreign_audit_file.criteria.len(), name);
}
```

src/storage.rs
```diff
@@ -1332,44 +1345,56 @@ fn parse_imported_criteria(value: toml::Value) -> Option<CriteriaEntry> {
...
+    } else {
+        Err(format!(
+            "no known criteria: {}",
+            audit_criteria
+                .iter()
+                .map(ToString::to_string)
+                .collect::<Vec<_>>()
+                .join(", "),
+        ))
     }
```
mystor, line 1370: Rather than doing the split like this, it might be cleaner (and help with situations where some but not all criteria are discarded) to log incrementally for each criteria removed, like this:

```rust
// Could be more elegant with `Vec::extract_if`, however that is still unstable
audit.criteria.retain(|criteria_name| {
    if !is_known_criteria(valid_criteria, criteria_name) {
        info!("discarding unknown criteria in imported audit: {criteria_name}");
        return false;
    }
    true
});
if audit.criteria.is_empty() {
    info!("imported audit parsing failed due to no known criteria");
    return None;
}
```

mystor (changes requested):

src/storage.rs
anforowicz: Thanks for catching this.  I didn't think about the built-in `safe-to-run` and `safe-to-deploy` criteria when adding this warning.

src/storage.rs
anforowicz: This is already done (almost, it doesn't include `audit_file.criteria.len()` as in your proposal, but it seems ok):

```
if !ignored_criteria.is_empty() {
    warn!(
        "Ignored {} invalid criteria entries when importing from '{}'\n\
        These criteria may have been made with a more recent version of cargo-vet",
        ignored_criteria.len(),
        name
    );
    info!(
        "The following criteria were ignored when importing from '{}': {:?}",
        name, ignored_criteria
    );
}
```

So, I went ahead and reverted the extra warnings from the beginning of `fn foreign_audit_file_to_local`.  The new warnings were based on a misunderstanding on my side - I've incorrectly thought that there needs to be at least one valid custom criteria definition (because I didn't think about `safe-to-deploy` and/or `safe-to-run`).

src/storage.rs
anforowicz: Done (I think).

anforowicz:
Thanks for reviewing @mystor.  Can you PTAL again?

src/storage.rs
```diff
@@ -1332,44 +1333,51 @@ fn parse_imported_criteria(value: toml::Value) -> Option<CriteriaEntry> {
...
+    if audit_criteria.is_empty() {
         info!("imported audit parsing failed due to no known criteria");
-        return None;
     }
+}
 
-    Some(audit)
+/// Parse an unparsed audit entry, validating and returning it.
+fn parse_imported_audit(valid_criteria: &[CriteriaName], value: toml::Value) -> Option<AuditEntry> {
+    let mut audit_entry = parse_from_value::<AuditEntry>(value).ok()?;
```
mystor, line 1358: This line dropped a previously-present log to tell us if imported audit parsing failed due to a serde parsing error, it'd be nice to have that back.

```suggestion
    let mut audit: AuditEntry = parse_from_value(value)
        .map_err(|err| info!("imported audit parsing failed due to {err}"))
        .ok()?;
```

src/storage.rs
```diff
@@ -1332,44 +1333,51 @@ fn parse_imported_criteria(value: toml::Value) -> Option<CriteriaEntry> {
...
 
 /// Parse an unparsed wildcard audit entry, validating and returning it.
 fn parse_imported_wildcard_audit(
     valid_criteria: &[CriteriaName],
     value: toml::Value,
 ) -> Option<WildcardEntry> {
-    let mut audit: WildcardEntry = parse_from_value(value)
-        .map_err(|err| info!("imported wildcard audit parsing failed due to {err}"))
-        .ok()?;
+    let mut wildcard_entry = parse_from_value::<WildcardEntry>(value).ok()?;
```
mystor, line 1373: This change means we no longer log errors from `parse_from_value` if the parsing failed due to a serde parse error - could we perhaps revert this change to get that log back?

```suggestion
    let mut audit: WildcardEntry = parse_from_value(value)
        .map_err(|err| info!("imported wildcard audit parsing failed due to {err}"))
        .ok()?;
```

src/storage.rs
```diff
@@ -1332,44 +1333,51 @@ fn parse_imported_criteria(value: toml::Value) -> Option<CriteriaEntry> {
...
+            info!("discarding unknown criteria in imported audit: {criteria_name}");
+            return false;
+        }
+        true
+    });
+
+    if audit_criteria.is_empty() {
         info!("imported audit parsing failed due to no known criteria");
-        return None;
     }
```
mystor, line 1353: This log feels a touch more appropriate to put in the caller, as theoretically we could retain only known criteria but not fail parsing the import (not that this ever happens). This change also means that it doesn't state whether it was an imported audit or wildcard audit.

src/storage.rs
```diff
@@ -1332,44 +1333,51 @@ fn parse_imported_criteria(value: toml::Value) -> Option<CriteriaEntry> {
...
+}
 
-    Some(audit)
+/// Parse an unparsed audit entry, validating and returning it.
+fn parse_imported_audit(valid_criteria: &[CriteriaName], value: toml::Value) -> Option<AuditEntry> {
+    let mut audit_entry = parse_from_value::<AuditEntry>(value).ok()?;
+
+    retain_only_known_criteria(&mut audit_entry.criteria, valid_criteria);
+    if audit_entry.criteria.is_empty() {
+        None
```
mystor, line 1362:
```suggestion
        info!("imported audit parsing failed due to no known criteria");
        None
```

src/storage.rs
```diff
@@ -1332,44 +1333,51 @@ fn parse_imported_criteria(value: toml::Value) -> Option<CriteriaEntry> {
...
 
-    audit
-        .criteria
-        .retain(|criteria_name| is_known_criteria(valid_criteria, criteria_name));
-    if audit.criteria.is_empty() {
-        info!("imported wildcard audit parsing failed due to no known criteria");
-        return None;
+    retain_only_known_criteria(&mut wildcard_entry.criteria, valid_criteria);
+    if wildcard_entry.criteria.is_empty() {
+        None
```
mystor, line 1377:
```suggestion
        info!("imported wildcard audit parsing failed due to no known criteria");
        None
```

mystor (approved):

src/storage.rs
anforowicz: Done - I inlined everything back (cause I guess distinguishing between audits and wildcard audits also matters in the other log message above).

anforowicz:
Thanks for the review @mystor!

Two quick notes:

* I have a separate PR to clean up Clippy and warnings: https://github.com/mozilla/cargo-vet/pull/598
* It seems to me that this PR passes `cargo test` on CI.  My local `cargo insta` state somehow got messed up and I see `cargo test` failures locally, but I assume that this is just happening on my machine and shouldn't block this PR.

mystor:
> It seems to me that this PR passes cargo test on CI. My local cargo insta state somehow got messed up and I see cargo test failures locally, but I assume that this is just happening on my machine and shouldn't block this PR.

This is actually #602 as far as I can tell, and those failures should be fixed now since #603.

mystor:
Thanks!

Issue #589, opened 2024-01-25
`cargo vet certify` unnecessarily reformats `config.toml`


anforowicz:
In my project the `supply-chain/config.toml` file is auto-generated from project-specific metadata by a tool (*).  The file has a copyright header and some other headers like `# @generated by tools/crates/gnrt vendor. Do not edit.`.  These headers are clobbered by some `cargo vet` commands.

I understand that `cargo vet fmt` will reformat all the files, including `config.toml`.  OTOH, IMO `cargo vet certify ...` should only write to `audits.toml` and should leave `config.toml` unchanged.  Is that a reasonable expectation?

------------------

(*) We use the `gnrt` tool to mapping project-specific categorization of crates into `cargo vet` requirements like:

```
[policy."autocfg:1.1.0"]                                               
criteria = ["does-not-implement-crypto", "safe-to-deploy", "ub-risk-2"]
```

bholley:
This seems reasonable at first glance. Certain certifications can necessitate changes in `config.toml`, but I think activating them entails `prune`? @mystor WDYT?

mystor:
`cargo vet certify` can cause changes to `config.toml`. For example, a new `certify` call will remove now-unnecessary exemptions for the crate which was just certified. This happens automatically for the crate which was certified in order to keep the tree relatively clean even without a manual `prune` call.

In general, all `cargo vet` commands which load the store also reformat it, even if there are no other changes to the store. When running with `--locked`, the `cargo vet` subcommand will actually [fail](https://github.com/mozilla/cargo-vet/blob/67e92b6d03f48802907b2c80119ed328289a114d/src/storage.rs#L664-L702) if the formatting of the store doesn't match the formatting which would be generated.

anforowicz:
Would it be possible to somehow split `config.toml` into 1) `cargo vet`-maintained/reformattable part and 2) human-authored part (in our case - with crate-specific, project-specific, separately-generated `criteria = ...` entries - an example can be found in the original issue report/comment above)?  (Just trying to brainstorm a way forward here.)

anforowicz:
[...]

PR #595, opened 2024-03-13
Avoid writing `config.toml` and other files if no significant changes.

anforowicz:
After this commit `audits.toml`, `config.toml`, and `imports.lock` will not be written if their contents wouldn't have changed ("contents" in a business logic sense - ignoring comments and/or formatting differences).

Fixes https://github.com/mozilla/cargo-vet/issues/589


anforowicz:
Ooops... I neglected to update test expectations... Let me fix this...

mystor (changes requested):
As I explain in https://github.com/mozilla/cargo-vet/issues/589#issuecomment-1998315628, I don't think this is the approach we want to take. Marking as requesting changes for now.

anforowicz:
Abandoning this PR based on https://github.com/mozilla/cargo-vet/issues/589#issuecomment-1998315628

Issue #484, opened 2023-05-29
Publish pre-built binaries of cargo-vet


repi:
While this tool is still in development, I do feel it and the ecosystem has matured quite a bit and would love to easily run it in CI in multiple of our repositories now. 

One can do it with building & caching it as suggested in the [docs](https://mozilla.github.io/cargo-vet/configuring-ci.html), but it would be much more preferable to have pre-built binaries published for main linux/windows/mac targets (esp. linux) directly in the releases here (for example like [cargo-deny](https://github.com/EmbarkStudios/cargo-deny/releases/tag/0.13.9) or [scache](https://github.com/mozilla/sccache/releases/tag/v0.5.0) does it). Would be faster CI and easier to integrate in solutions that do not have caching.

Would this be possible to setup? One can take inspiration from some existing repos (like mentioned above) of how they build and publish it through github actions.

bholley:
Would be happy to take a patch for this.

repi:
this could be very good use case for the new [`cargo-dist`](https://github.com/axodotdev/cargo-dist) to set up release workflow and deployment of binaries and try it out (and give feedback to @axodotdev about it).

maybe something @Gankra at axo would be interested in as another real test project to use it on?

staktrace:
I have a PR up at https://github.com/mozilla/cargo-vet/pull/600 that sets up the `cargo-dist` release workflow. It was pretty painless, kudos @Gankra (and team) for that.

For context, we'd like prebuilt binaries published so that we can use them via the [hermit](https://github.com/cashapp/hermit) tool.

PR #599, opened 2024-03-27
Update metadata for new home

staktrace:

PR #600, opened 2024-03-27
Install cargo-dist to generate releases

staktrace:
This PR was generated by running `cargo dist init` and selecting all the default options. It seems to have pretty sane behaviour out of the box.

This requires the repository URL fix in https://github.com/mozilla/cargo-vet/pull/599 to work correctly.

Fixes https://github.com/mozilla/cargo-vet/issues/484

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,266 @@
+# Copyright 2022-2023, axodotdev
+# SPDX-License-Identifier: MIT or Apache-2.0
+#
+# CI that:
+#
+# * checks for a Git Tag that looks like a release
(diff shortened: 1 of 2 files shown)


staktrace:
For anybody following along on this, https://matrix.to/#/!dRHkXGXAEhtBOpnLTY:mozilla.org/$4_H4pbu-FftonOnqfHJLXWZzd1ZNHSDdtptBk1IW_r4?via=mozilla.org

bholley (approved):
LGTM. I only skimmed the logic but I understand this is vanilla output from cargo-dist, which was built by the same person who built the initial version of cargo-vet.

PR #609, opened 2024-05-12
remove dollar sign from shell commands

GlenDC:
Copying these commands otherwise do not work out of the box. I get the doc convention to vizually indicicate a command over output, but it interfers with the copy aspect.


bholley:
Thanks! There are a few other places in the docs that use dollar signs, but these are probably the ones most likely to be copied.

bholley (approved):

PR #614, opened 2024-05-24
Include criteria which imply the required criteria in suggest output

### LINKED ISSUE #613: Clarify which audits are recommended when `X` `implies` `Y` criteria
## Context

In my experience, engineers performing semi-automated crate updates are sometimes confused which criteria to audit for.  For example, an engineer may see the following message from `cargo vet`:

```
$ tools/crates/create_update_cl.py --no-upload single skrifa
...Updating skrifa@0.19.0 to a newer version...
$ tools/crates/run_cargo_vet.py check
...Vetting Failed!

1 unvetted dependencies:
  skrifa:0.19.1 missing ["safe-to-deploy", "crypto-safe", "ub-risk-2"]

recommended audits for safe-to-deploy, crypto-safe, ub-risk-2:
    Command                              Publisher  Used By   Audit Size
    cargo vet diff skrifa 0.19.0 0.19.1  rsheeter   chromium  12 files changed, 1317 insertions(+), 220 deletions(-)

estimated audit backlog: 1537 lines

```

The crate [has been previously audited](https://source.chromium.org/chromium/chromium/src/+/main:third_party/rust/chromium_crates_io/supply-chain/audits.toml;l=1171-1182;drc=d89e1c184ad53a7b09cf6f69f45a16e8b5695c98) as `does-not-implement-crypto` and `ub-risk-0` and the delta doesn't change those properties.  But `cargo vet`'s message recommends auditing for `crypto-safe` and `ub-risk-2` instead (which have a transitive `implies` relationship: `ub-risk-0` implies `ub-risk-1` which implies `ub-risk-2`; and `does-not-implement-crypto` implies `crypto-safe` - see [here](https://source.chromium.org/chromium/chromium/src/+/main:third_party/rust/chromium_crates_io/supply-chain/audits.toml;l=23-39;drc=d89e1c184ad53a7b09cf6f69f45a16e8b5695c98) and [here](https://source.chromium.org/chromium/chromium/src/+/main:third_party/rust/chromium_crates_io/supply-chain/audits.toml;l=21;drc=d89e1c184ad53a7b09cf6f69f45a16e8b5695c98)).

To reduce the confusion, we've tried to cover this in our documentation:

* https://github.com/google/rust-crate-audits/blob/main/auditing_standards.md#audits-should-be-as-precise-as-possible
* [`tools/crates/create_update_cl.md`](https://chromium-review.googlesource.com/c/chromium/src/+/5558983/2/tools/crates/create_update_cl.md):

    ```
    - Also note that `cargo vet` will list the _minimum_ required criteria
      and `audits.toml` can and should record stricter certification if
      possible. In particular:
         - Record `does-not-implement-crypto` instead of `crypto-safe` if the
           crate does not implement crypto.
         - Record a lower-numbered `ub-risk-N` if appropriate.    
    ```

But, it would be great if `cargo vet`'s output could also somehow be tweaked to avoid this kind of confusion (since the doc guidance is easy to miss, but the output of `cargo vet` is looked at each time a crate is updated).  Can we discuss if and how `cargo vet`'s output could be changed?

FWIW this confusion (in light of `does-not-implement-crypto` vs `crypto-safe`) has also been touched upon in https://github.com/mozilla/cargo-vet/issues/417 but that other issue focuses more on delta audit workflow surfacing the properties of the baseline audit.

## Concrete proposal

I would like to propose changing this part of the output:

    ```
    recommended audits for safe-to-deploy, crypto-safe, ub-risk-2:
    ```

Into:

    ```
    recommended audits for safe-to-deploy, crypto-safe (or does-not-implement-crypto), ub-risk-2 (or ub-risk-0, or ub-risk 1):
    ```

It seems possible to derive the new text based on the `implies` relationship that `cargo vet` already recognizes between the audit criteria.

I haven't yet tried looking at `cargo vet`'s source code to see where such change could be made.  Obviously we wouldn't want to change all criteria => string conversions - hopefully it won't be too difficult to identify which ones can use the new, expanded wording (?).

WDYT?

mystor:
For most consumers using exclusively the default criteria, this will only impact the output for `safe-to-run` recommendations to mention that `safe-to-deploy` would also be applicable. For consumers using more complex criteria, this may increase the size of the suggest output more substantially.

This change does impact JSON output, but only for strings generally intended for human consumption. If a script was previously parsing these strings it may be broken by the change.

Fixes #613

### DIFF
--- a/src/criteria.rs
+++ b/src/criteria.rs
@@ -177,6 +177,16 @@ impl CriteriaMapper {
     pub fn criteria_index(&self, criteria_name: CriteriaStr<'_>) -> usize {
         self.index[criteria_name]
     }
+
+    /// Yields the indicies for all criteria which imply the the given criteria.
+    pub fn implied_by_indices(&self, criteria_idx: usize) -> impl Iterator<Item = usize> + '_ {
+        self.all_criteria_iter()
+            .enumerate()
+            .filter(move |&(idx, implies_set)| {
+                implies_set.has_criteria(criteria_idx) && criteria_idx != idx
+            })
+            .map(|(idx, _)| idx)
+    }
 }
 
 impl CriteriaSet {
--- a/src/criteria.rs
+++ b/src/criteria.rs
@@ -178,7 +178,7 @@ impl CriteriaMapper {
         self.index[criteria_name]
     }
 
-    /// Yields the indicies for all criteria which imply the the given criteria.
+    /// Yields the indices for all criteria which imply the given criteria.
     pub fn implied_by_indices(&self, criteria_idx: usize) -> impl Iterator<Item = usize> + '_ {
         self.all_criteria_iter()
             .enumerate()
--- a/src/resolver.rs
+++ b/src/resolver.rs
@@ -1979,9 +1979,26 @@ impl<'a> ResolveReport<'a> {
 
         let mut suggestions_by_criteria = SortedMap::<CriteriaName, Vec<SuggestItem>>::new();
         for s in suggestions.clone().into_iter() {
+            // Generate a suggestion for which criteria to use for the given
+            // suggestion. For each criteria, also list out others which would
+            // imply the required criteria to surface the full set of options.
             let criteria_names = self
                 .criteria_mapper
-                .criteria_names(&s.suggested_criteria)
+                .minimal_indices(&s.suggested_criteria)
+                .map(|criteria_idx| {
+                    let name = self.criteria_mapper.criteria_name(criteria_idx);
+                    let implied_by = self
+                        .criteria_mapper
+                        .implied_by_indices(criteria_idx)
+                        .map(|idx| format!("or {}", self.criteria_mapper.criteria_name(idx)))
+                        .collect::<Vec<_>>();
+
+                    if implied_by.is_empty() {
+                        name.to_owned()
+                    } else {
+                        format!("{} ({})", name, implied_by.join(", "))
+                    }
+                })
                 .collect::<Vec<_>>()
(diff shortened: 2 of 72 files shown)


src/criteria.rs
```diff
@@ -177,6 +177,16 @@ impl CriteriaMapper {
     pub fn criteria_index(&self, criteria_name: CriteriaStr<'_>) -> usize {
         self.index[criteria_name]
     }
+
+    /// Yields the indicies for all criteria which imply the the given criteria.
```
anforowicz, line 181: optional nitty nit: s/the the/the/

anforowicz (approved):

anforowicz:
Thanks!  LGTM (although I am not sure if it counts because I am not a real owner / reviewer of this project).

src/criteria.rs
afranchuk, line 181: another nit (suggestion aggregated with @anforowicz's comment)
```suggestion
    /// Yields the indices for all criteria which imply the given criteria.
```

afranchuk (approved):
This looks good.

I'd like to bring up a complaint in the human-readable output that I've had prior to this PR. We say "recommended audits for safe-to-run (or safe-to-deploy):", however to me that sentence structure makes it sound like those are audits you need to do to meet some "safe-to-run" objective overall. This doesn't make sense in context since you should be told the criteria which the audits must meet, however it still throws me off when I read it. I think it's clearer if the wording is "recommended safe-to-run (or safe-to-deploy) audits:" or "recommended audits {meeting,satisfying} safe-to-run (or safe-to-deploy):".

PR #615, opened 2024-06-02
Added support for Cargo `http.cainfo`

ranger-ross:
This PR adds support for using a custom CA via the Cargo [http.cainfo](https://doc.rust-lang.org/cargo/reference/config.html#httpcainfo) configuration.

See https://github.com/mozilla/cargo-vet/issues/612 for more info


mystor (changes requested):
I'm not sure we want to add a hard dependency on the full `cargo` crate. That's a pretty heavy dependency to pull in just to read the cargo config file.

The API of the `cargo` crate is also quite unstable (as noted in the [documentation](https://docs.rs/cargo/0.80.0/cargo/index.html), in fact the latest version appears to have a different API than the 0.65.0 which you're depending on here).

I'm open to us using cargo HTTP configuration options to configure our reqwest client. Ideally we'd shell out to the `cargo` subcommand with `cargo config get`, though that feature is unfortunately still unstable (https://github.com/rust-lang/cargo/issues/9301).

For now, to support this unfortunately I think we'd need to parse the cargo config options independently, rather than relying on the `cargo` crate to do so. Based on the [cargo configuration documentation](https://doc.rust-lang.org/cargo/reference/config.html#hierarchical-structure), the process would roughly be to try to parse TOML from each `$ANCESTOR_DIR/.cargo/config` or `$ANCESTOR_DIR/.cargo/config.toml`, then try the next ancestor up, followed by finally checking `$CARGO_HOME/config[.toml]`. There's also a `CARGO_HTTP_CAINFO` environment variable which would need to be checked.

ranger-ross:
Hey @mystor , 
Thanks, thats a totally fair response. (I was also not in love with taking a dependency with cargo)

I updated this PR to remove the dependency on cargo and self implemented the logic for resolving cargo configurations by following the documentation [here](https://doc.rust-lang.org/cargo/reference/config.html).

Let me know what you think :+1:

mystor (approved):
I applied some clean-ups and audited the new dependencies. Approving with those changes.

Thanks for the patch!

PR #621, opened 2024-07-09
Always prune imports when importing new audits for a crate

### LINKED ISSUE #620: Consider pruning publisher entries even if `prune_imports` is disabled when a new publisher entry is being added for a crate
In order to avoid unnecessary changes to files such as `imports.lock` cargo-vet makes some effort to avoid removing entries from the file unless requested. The main way this happens in normal operation is when `cargo vet certify` is used to add a new audit for a crate, which can lead to other exemptions, imports, etc. for the crate to be removed.

Unfortunately this pattern doesn't work super well for crates with wildcard audits or trusted entries. For these crates, generally adding a dependency does not require the user running any command other than `cargo vet`, which does not prune unused entries, which means that the previous `publisher` entries are left behind and not cleaned up.

We could consider automatically pruning `publisher` entries for crates even when `prune_imports` is disabled so long as a new publisher entry is being added for that crate. This still avoids unnecessary updates, but means we can automatically clean up publisher entries when new entries are being added anyway.

We could also consider doing this for other forms of imports such as normal and wildcard audits, though removing those from `imports.lock` due to them becoming unnecessary is less frequent during normal dependency updates.

mystor:
Currently we only prune imports when doing an explicit  prune or when certifying a new audit for the crate. This patch updates the behaviour to also prune imports whenever we are otherwise also adding new imports for the given crate.

This gives us an opportunity to automatically prune old publisher entries when updating a dependency with a trusted entry or wildcard audit, which otherwise wouldn't happen until `cargo vet prune` is explicitly called.

This pruning happens after resolution, so existing imports will still be preferred, and `cargo vet prune` may still perform additional cleanup. Exemptions, git audits, and unpublished entries aren't impacted.

Fixes #620

### DIFF
--- a/src/resolver.rs
+++ b/src/resolver.rs
@@ -2858,6 +2858,39 @@ pub fn update_store(
     get_store_updates(cfg, store, mode).apply(store);
 }
 
+/// Helper function to determine if we should be pruning imports.
+///
+/// We always prune if requested, but will also prune imports if a new audit or
+/// publisher entry is being added for the crate.
+fn should_prune_imports(
+    store: &Store,
+    required_entries: &Option<SortedMap<RequiredEntry, CriteriaSet>>,
+    mode: UpdateMode,
+    pkgname: PackageStr<'_>,
+) -> bool {
+    if mode.prune_imports {
+        true
+    } else if let Some(required_entries) = required_entries {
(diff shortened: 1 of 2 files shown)


afranchuk (approved):
Looks good, I only had one small suggestion.

PR #622, opened 2024-07-10
Update cargo-dist to version 0.18.0

mystor:


.github/workflows/release.yml
```diff
@@ -246,21 +279,3 @@ jobs:
...
-          # Remove the granular manifests
-          rm -f artifacts/*-dist-manifest.json
-      - name: Create Github Release
-        uses: ncipollo/release-action@v1
-        with:
-          tag: ${{ needs.plan.outputs.tag }}
-          name: ${{ fromJson(needs.host.outputs.val).announcement_title }}
-          body: ${{ fromJson(needs.host.outputs.val).announcement_github_body }}
-          prerelease: ${{ fromJson(needs.host.outputs.val).announcement_is_prerelease }}
-          artifacts: "artifacts/*"
```
afranchuk, line 266: I wonder whether the `announce` job serves any purpose anymore...

afranchuk (approved):

.github/workflows/release.yml
mystor: I'm not sure, it seems like `cargo-dist` still generates it at least. Looks like it's used at least for certain publish jobs like `homebrew` formulas: https://github.com/axodotdev/cargo-dist/blob/6439c4774e5011c88729044eecd90b08361997b7/cargo-dist/templates/ci/github/release.yml.j2#L526-L541

Looks like the change happened quite recently in https://github.com/axodotdev/cargo-dist/commit/4b8b18d315a92d1c4aac3063ff01acd0a5e4b250, where they changed around when the github release happens, and now by default it looks like `announce` is empty - perhaps they'll remove it from the template in a later version.

Issue #380, opened 2023-01-19
Consider sub-crate feature exclusions for audits


bholley:
Generally speaking cargo-vet always operates at crate granularity, and we've resisted anything finer-grained on the grounds that it's difficult to reason about, difficult to implement, and of questionable utility in practice.

Earlier today, @selenography was describing his in-progress audit of `reqwest`, and asking whether there was a way to exclude the `socks` feature, which was a lot of unfamiliar code which we aren't using anyway. This led @mystor to formulate a feature description that I thought was interesting enough to at least document and discuss further:

The key insight is that `cargo-vet` already assumes that any code not reachable by enabling all features on the top-level crates (i.e., passing `--all-features` to `cargo-metadata`) can be excluded from analysis. At the moment this happens by dropping conditional dependencies which can never be built, because that's what cargo-vet knows how to do. However, we could also apply this same conditionality to the inclusion/exclusion of audits.

A rough sketch of this would be an optional `exclude-features = [...]` field on audit entries. Cargo-vet would only consider the audit if it were able to verify that none of the features in the list could ever be enabled on that node in the crate graph. Getting this information out of `cargo-metadata` might be anywhere between trivial and impossible, I haven't investigated.

You could then imagine an additional `add-features = [...]` field, which would allow you to patch in coverage to an existing audit. So if @selenography audited `reqwest-x.y.z` without `socks`, and somebody comes along and wants to use `socks`, they could just audit the `socks` part, and cargo-vet could combine the two.

mystor:
In terms of how difficult `exclude-features` is to implement, I think it should be fairly straightforward. Cargo doesn't support disabling a feature in a dependency when a feature is enabled (features are additive) so when we run `cargo metadata` with `--all-features`, we're seeing the maximal crate graph from the metadata command. We already have the list of enabled features for each package node in the graph (https://docs.rs/cargo_metadata/latest/cargo_metadata/struct.Node.html#structfield.features), so we could effectively disable an audit edge with `exclude-features = [...]` if any of the listed features are present in the given node.

This has a slightly awkward interaction with the evolution of crates over time, as a given feature could e.g. no longer exist in a newer version (as all checks are done against the current version in the dependency graph, as we don't know e.g. "what features would be enabled on a previous version X of this crate"), or the feature could be renamed.

[...]

PR #623, opened 2024-07-10
Add SecureDrop's aggregated audits to the registry

legoktm:
From <https://github.com/freedomofpress/securedrop-supply-chain>.

-- 

We weren't sure if "securedrop" or "freedomofpress" would be preferred; so far all of our audits are specifically for SecureDrop and not other FPF projects.

### DIFF
--- a/registry.toml
+++ b/registry.toml
@@ -16,5 +16,8 @@ url = "https://raw.githubusercontent.com/divviup/libprio-rs/main/supply-chain/au
 [registry.mozilla]
 url = "https://raw.githubusercontent.com/mozilla/supply-chain/main/audits.toml"
 
+[registry.securedrop]
+url = "https://raw.githubusercontent.com/freedomofpress/securedrop-supply-chain/main/audits.toml"
+
 [registry.zcash]
 url = "https://raw.githubusercontent.com/zcash/rust-ecosystem/main/supply-chain/audits.toml"


bholley:
Actually, @legoktm , can you confirm all of these audits correspond to full reviews of the relevant crates? Some of the notes indicate that they may be trusting based on source (e.g., "Rust project member"), in which case `trusted` declarations would be more appropriate.

legoktm:
hi @bholley, I double checked and the "Rust Project member" comments only apply to `[trusted]` entries - let me know of course if we're doing that wrong :)

mystor:
I did a quick scan over your audits, and noticed a few of them have notes mentioning parts of the crate were skipped when auditing (e.g. [1](https://github.com/freedomofpress/securedrop-supply-chain/blob/3861fae563980ef60680b80ce216e06b04af1aa9/audits.toml#L12), [2](https://github.com/freedomofpress/securedrop-supply-chain/blob/3861fae563980ef60680b80ce216e06b04af1aa9/audits.toml#L122), [3](https://github.com/freedomofpress/securedrop-supply-chain/blob/3861fae563980ef60680b80ce216e06b04af1aa9/audits.toml#L242), [4](https://github.com/freedomofpress/securedrop-supply-chain/blob/3861fae563980ef60680b80ce216e06b04af1aa9/audits.toml#L385), [5](https://github.com/freedomofpress/securedrop-supply-chain/blob/3861fae563980ef60680b80ce216e06b04af1aa9/audits.toml#L392)). Published audits should always apply to all code in the crate for all targets.

Perhaps that could be written out more explicitly in the `safe-to-deploy` built-in criteria description.

legoktm:
> Published audits should always apply to all code in the crate for all targets.

Got it, I think all of these are my own audits so I'll spend a bit of time reviewing the missing bits and updating those audits and then come back here.

Relatedly, I think https://github.com/mozilla/cargo-vet/issues/380 is what we're really looking for.

mystor:
Closing this PR for now, but please feel free to re-open it once you've gotten around to updating those audits :-)

Issue #624, opened 2024-07-11
Add diff.rs as an alternate mode for `cargo vet diff`


divergentdave:
[diff.rs](https://diff.rs/) provides a single-page app to diff two `.crate` files, fetched from `crates.io`. Given that Sourcegraph has been unstable lately, I think it would be useful to add this as an alternative to `--mode=local` and `--mode=sourcegraph`. Note that this only shows diffs, so it wouldn't be appropriate for `cargo vet inspect`.

mystor:
This has been fixed in #625

PR #625, opened 2024-07-11
Add support for diff.rs to `cargo vet diff`

divergentdave:
This implements #624.


mystor (approved):
Thanks! This seems like a useful tool to have as an alternative

PR #630, opened 2024-09-03
More consistently tidy files before serializing

mystor:
Before these changes, we wouldn't tidy the files when aggregating, which could lead to empty entries appearing in aggregated audits files (e.g. github.com/mozilla/supply-chain/blob/0473560de259e980a703711cfbe64987c22fe3df/audits.toml#L838)

### DIFF
--- a/src/format.rs
+++ b/src/format.rs
@@ -2,7 +2,7 @@
 
 use crate::errors::{StoreVersionParseError, VersionParseError};
 use crate::resolver::{DiffRecommendation, ViolationConflict};
-use crate::serialization::spanned::Spanned;
+use crate::serialization::{spanned::Spanned, Tidyable};
 use crate::{flock::Filesystem, serialization};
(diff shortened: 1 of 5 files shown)


src/format.rs
```diff
@@ -235,6 +235,23 @@ pub struct AuditsFile {
     pub trusted: TrustedPackages,
 }
 
+impl Tidyable for AuditsFile {
+    fn tidy(&mut self) {
+        self.audits.retain(|_, entries| {
+            entries.sort();
+            !entries.is_empty()
```
afranchuk, line 242: Due to the repetition, I wonder whether it'd make sense to have `impl<K, V: Ord> Tidyable for SortedMap<K, Vec<V>>` and use that.

afranchuk (approved):

mystor:
I believe this CI failure will be fixed by #632

Issue #629, opened 2024-08-15
"does not exist" error for existing package


sophie-h:
The following works for many packages but not for `librsvg-rebind`.

Maybe it's an issue with the package only having one release? Or is something wrong with the package? If it's the latter, a better error message would be nice, if possible.

```
$ cargo vet certify --notes "GNOME contributor" --accept-all --criteria safe-to-deploy --wildcard federicomenaquintero librsvg-rebind
ERROR   × Cannot fetch crate information, 'librsvg-rebind' does not exist.
```

mystor:
In order to reduce the load on crates.io's APIs (see https://crates.io/data-access#api), we attempt to do some local caching of crate metadata, such as publisher information. As part of that, though, we also cache information about whether or not a crate exists. 

I'm guessing that your local cache ended up seeing at some point in the past that `librsvg-rebind` was not published, and cached that information, and the code is failing to invalidate the cache right now.

Theoretically with the sparse `index.crates.io` index now, and that being explicitly not-rate-limited, we could avoid needing any kind of cache for that basic version information, and only use the cache for the rate-limited metadata information, which would help avoid these kinds of outdated-cache issues.

sophie-h:
Thanks a lot! I manually deleted the entries from the cache and it worked.

If your suggestion is not implemented, maybe enhancing the error message would be good. Or adding an option to clean the cache just for the added crate.

PR #631, opened 2024-09-03
Simplify crates.io API caching

mystor:
Previously, we would perform time-based caching for data from the crates.io API in order to avoid unnecessary load on it. Unfortunately, this would sometimes lead to situations such as #629 where the cache is out-of-date and confusing errors are produced.

Since the last time we looked at this, the API guidance for the sparse HTTP index has been clarified, explicitly having "no rate limit". This allows us to simplify the caching code dramatically, caching only the data from the rate-limited crates.io API, and always checking the sparse HTTP index to see if new versions have been published.

The new caching system no longer needs expiry times, as it will be updated consistently whenever a new version is published to the crates.io index.

After this change, we will make more network requests than before when running cargo-vet to check the index.


mystor:
Note that this change also starts collecting the sha256 hashes for versions when fetching from the index. I haven't implemented actually validating these when downloading yet (though that should be done at some point, and likely won't be too difficult, beyond requiring adding & auditing new dependencies).

src/storage.rs
```diff
@@ -2333,429 +2291,218 @@ impl Cache {
...
-    cache: &'a Cache,
-    crate_name: PackageStr<'a>,
-    cache_expiration: Option<chrono::Duration>,
-    versions: Option<FastSet<&'a semver::Version>>,
-    need_version_details: bool,
-    need_crate_metadata: bool,
-}
+                let mut result = SortedMap::new();
+                for version_entry in crate_entry.versions() {
+                    if let Ok(version) = semver::Version::parse(version_entry.version()) {
```
afranchuk, line 2359: Maybe log this error? It should be unlikely, though.

afranchuk (approved):
Looks pretty good overall!

Issue #611, opened 2024-05-22
SourceGraph is unreliable


cemoktra:
SourceGraph is getting unreliable, e.g.:
https://sourcegraph.com/crates/prost-build/-/compare/v0.12.3...v0.12.6?visible=7000

This ends in "Empty Repository" today with reason being something like "severely out of date". While the alternative local diff exists, the diff is much harder to read (no color highlighting etc).

It could be an alternative to support configuring local UI tools

afranchuk:
I've experienced this with about 8 diffs as well. For example, https://sourcegraph.com/crates/gimli/-/compare/v0.28.1...v0.29.0. I have not had a diff succeed, and resorted to using local tools.

jdorfman:
Hi @afranchuk looking into it

bholley:
Had a good chat with @beyang today. Sourcegraph is currently upgrading and streamlining the infrastructure for the sourcegraph.com instance, which should ultimately make the crate indexing more reliable but is resulting in some hiccups in the mean time. They expect it'll be resolved by the end of the month. In the mean time, folks can pass `--local` for crates that don't resolve on Sourcegraph.

PR #633, opened 2024-09-18
Remember user --mode choices, and support diff.rs for inspect

mystor:
This change adds support to remember user's --mode choices, so that they do not need to be used every time. In addition, it adds a basic support for diff.rs to inspect (by loading a diff from a version to itself) such that this mode memory can be shared between diff and inspect.

If xfbs/diff.rs#24 is merged, we should change the diff.rs inspect behaviour to use the new browse endpoints instead.

Finally, this patch changes the default mode from sourcegraph to diff.rs. As noted in #611, sourcegraph has been quite unreliable lately, and diff.rs is likely to provide a better experience until that is resolved.

### DIFF
--- a/src/storage.rs
+++ b/src/storage.rs
@@ -19,6 +19,7 @@ use tar::Archive;
 use tracing::{error, info, log::warn, trace};
 
 use crate::{
+    cli::FetchMode,
     criteria::CriteriaMapper,
     errors::{
         AggregateError, BadFormatError, BadWildcardEndDateError, CacheAcquireError,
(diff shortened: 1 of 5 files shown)


src/storage.rs
```diff
@@ -2291,6 +2292,16 @@ impl Cache {
...
+    pub fn get_last_fetch_mode(&self) -> Option<FetchMode> {
+        let guard = self.state.lock().unwrap();
+        guard.command_history.last_fetch_mode
+    }
+
+    pub fn set_last_fetch_mode(&self, mode: FetchMode) {
+        let mut guard = self.state.lock().unwrap();
+        guard.command_history.last_fetch_mode = Some(mode);
+    }
+
```
afranchuk, line 2304: I'd like to see (either in addition to, or replacing these in the `pub` interface):
```rust
pub fn fetch_mode(&self, chosen_mode: Option<FetchMode>) -> FetchMode {
    if let Some(mode) = chosen_mode {
        self.set_last_fetch_mode(mode);
    }
    self.get_last_fetch_mode().unwrap_or(FetchMode::DiffRs)
}
```
This deduplicates the logic, however my main reasoning is to have one place where the default `FetchMode` is established. And/or we could add a `impl Default for FetchMode`.

afranchuk (approved):

src/storage.rs
mystor: I centralized & documented the logic a bit more, also moving the local-forcing stuff into the function a bit to make it more explicit.

PR #636, opened 2024-10-03
Bump to 0.10.0

mystor:


afranchuk (approved):

chris-morgan:
Is there any sort of changelog? I reckon there should be for all projects, and *especially* for something like this, but I can’t find anything (no file, nothing in the tag or release, nothing in here).

mystor:
I've filed #639 to actually add a changelog. Not sure why we didn't have one before now.

PR #637, opened 2024-10-14
Add Actix audit file to registry

robjtede:
I've started an audit log for all Actix first-party and some third-party ecosystem crates.

### DIFF
--- a/registry.toml
+++ b/registry.toml
@@ -1,3 +1,6 @@
+[registry.actix]
+url = "https://raw.githubusercontent.com/actix/supply-chain/main/audits.toml"
+
 [registry.bytecode-alliance]
 url = "https://raw.githubusercontent.com/bytecodealliance/wasmtime/main/supply-chain/audits.toml"
 


mystor (changes requested):
Thanks!

I notice your audits.toml file has [inline comments](https://github.com/actix/supply-chain/blob/7f5f3473acea3d0c6e9abda2056b7446b24180b9/audits.toml#L5-L7), suggesting it is manually managed, rather than created using `cargo-vet`. Are you aware of the stability requirements of an importable audits.toml file (e.g. never removing an audit from the file) which would normally be managed by `cargo vet` if it was used to generate the file?

I notice that some audits were removed from the file in previous changes (such as [this one](https://github.com/actix/supply-chain/commit/22c7f7441a4d84ac69cfe050b7302612e5bb6a19)). Changes like this could cause vet to start failing for folks who are depending on your published audits.

robjtede:
> manually managed

Totally get the concern and happy to recreate this file purely with the cargo vet commands before telling more people about it.

> some audits were removed from the file in previous changes

yes, i'd learned about wildcard-audits and that seemed like an obvious improvement for the self-certify cases

> stability requirements of an importable audits.toml file (e.g. never removing an audit from the file)

yes, very reasonable requirement and will certainly uphold it going forwards

mystor (approved):
Thanks! Just wanted to check in about those things given the manually curated audits.toml file before adding actix to the registry :-)

PR #639, opened 2024-10-18
Add CHANGELOG.md to track changes between cargo-vet releases

mystor:


afranchuk (approved):
What's nice is, I believe, cargo-dist will handle updating the ChangeLog headers (if I read the docs correctly). In other projects I've used `cargo-release` and it would do it, though it did require some manual configuration. I *think* cargo-dist will Just Work.

Issue #336, opened 2022-09-30
Basic auth and/or ssh url support for cargo aggregate


agraven:
Using cargo-vet to aggregate audit sources from private repositories can be quite difficult, as there's no way of accessing an audit file from a private source without doing something like putting basic auth information in the URL itself, which will then result in the username:password pair being spread all over in the form of `aggregated-from` keys.

The main approaches to addressing this use case I see are:
1. A command line flag (or two) for specifying basic auth credentials. (Or maybe they could be read from stdin?)
2. Support fetching audit files via ssh, which would delegate the authentication to the ssh-agent

What your thoughts on this? I'd be happy to give writing a PR for this functionality a go

bholley:
SSH sounds a little cleaner to me, though I'm not sure whether this would involve pulling in a bunch of dependencies or if we can lean on what cargo already does. @mystor WDYT?

mystor:
I don't believe we currently depend on libssh2, which would probably be the easiest way to fetch something with ssh. From a quick check locally this would pull in: `ssh2`, `libssh2-sys`, `lock_api`, `openssl-sys`, `parking_lot`, `parking_lot_core`, and `scopeguard`. 

We could also probably do something without these dependencies if we're fine with just invoking `scp` on the command line and using temporary files.

bholley:
Generally speaking I'd rather not take all those dependencies. I think I'm fine with either a command-line scp solution or an HTTP basic auth solution.

dexterduck:
In the interim, how about adding support for the `file://` URL scheme? 

As far as I can tell it's not possible to point `cargo vet aggregate` or the `imports` table in `config.toml` to a local file (please let me know if I'm missing something), which would make it easier to create an external workaround for pulling from private repositories.

I'd be happy to put a PR in if there's interest in the change.

mystor:
Support just for file URIs would likely be quite difficult to use, as those URIs are absolute, which isn't great for paths in a repository due to it not being cloned in a consistent location. The solution for local paths would probably need to support relative paths, which would involve tracking things like base URIs for what resources we're loading, which is a larger change.

I think the easiest way forward here is probably a PR adding support for the `ssh://` protocol by delegating to `scp`. This would require adding support to the methods in https://github.com/mozilla/cargo-vet/blob/56ebfded2d27d6c28a56832aaae1dc99b41c4311/src/network.rs.

bachp:
[...]

PR #640, opened 2024-10-24
Add support for authorization via git-credential helper

bachp:
This can for example be used to fetch audits from private/internal gitlab instances via the API.

Example:

1. Setup git credential helper for your website as documented in the git documentation: https://git-scm.com/docs/gitcredentials
2. Add an API URL to the `sources.list`: https://gitlab.com/api/v4/projects/12000/repository/files/supply-chain%2Faudits.toml/raw?ref=main
3. Run: `cargo vet aggregate --use-git-credentials --output-file audits.toml sources.list`

The reason for using git-credential helpers instead of .netrc is that it also supports authentication methods other than basic auth.

Addresses the HTTP authentication part of https://github.com/mozilla/cargo-vet/issues/336

### DIFF
--- a/src/credentials.rs
+++ b/src/credentials.rs
@@ -1,0 +1,274 @@
+//! Git credential helper integration
+//!
+//! This module provides integration with git credential helpers as documented in
+//! https://git-scm.com/docs/git-credential and https://git-scm.com/docs/api-credentials
+//!
+//! It supports both traditional username/password authentication and the newer
+//! authtype/credential format for more advanced authentication schemes.
+
+use reqwest::Url;
+use std::process::Stdio;
+use tokio::io::AsyncWriteExt;
+use tracing::debug;
+
+/// Represents credential information returned by git credential helper
+#[derive(Debug, Clone)]
+pub enum GitCredential {
+    /// Username and password for basic authentication
+    UsernamePassword { username: String, password: String },
+    /// Pre-formatted credential with authentication type
+    Formatted {
+        auth_type: String,
+        credential: String,
+    },
+}
+
+impl GitCredential {
+    /// Apply the credential to a reqwest RequestBuilder
+    pub fn apply_to_request(
+        self,
+        request_builder: reqwest::RequestBuilder,
+    ) -> reqwest::RequestBuilder {
+        match self {
+            GitCredential::UsernamePassword { username, password } => {
+                request_builder.basic_auth(username, Some(password))
+            }
+            GitCredential::Formatted {
+                auth_type,
+                credential,
+            } => request_builder.header("Authorization", format!("{} {}", auth_type, credential)),
+        }
+    }
+}
+
+/// Execute git credential fill for the given input
+///
+/// This function is separated to allow for easier testing by mocking the git call.
+async fn execute_git_credential_fill(
+    credential_input: &str,
+) -> Result<Option<String>, Box<dyn std::error::Error>> {
+    // Execute git credential fill
+    let mut child = tokio::process::Command::new("git")
+        .args(&["credential", "fill"])
+        .stdin(Stdio::piped())
+        .stdout(Stdio::piped())
+        .stderr(Stdio::piped())
+        .spawn()?;
+
+    // Write the credential input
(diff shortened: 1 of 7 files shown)


src/network.rs
```diff
@@ -284,9 +285,15 @@ impl Network {
...
-        let res = self
-            .client
-            .get(url.clone())
+        let mut request_builder = self.client.get(url.clone());
+
+        // Add a header from the environment, if present.
+        if let Ok(header) = std::env::var("CARGO_VET_AUTH_HEADER") {
+            let (name, value) = header.split_once(':').expect("Invalid header format");
+            request_builder = request_builder.header(name, value);
+        }
```
mystor, line 294: I think I'd prefer to avoid reading the environment every time we build a request. It might be nice to cache these headers once when we first build the `client`, and then re-use them for future requests.

I do worry a bit about this being an inflexible option as this would send the header to every single server which `cargo-vet` connects to, which may not be what folks want (e.g. you probably don't want to be sending your auth headers to `index.crates.io`?).

Something like adding an option to pass in a `--netrc-file` which should be used by `vet aggregate` seems perhaps like a more flexible solution which avodis the problem of sending the auth headers to every server you connect to, though admittedly it is more complex and would require adding a small netrc parser. It does appear that there are various crates which already exist to do this.

mystor (changes requested):

src/network.rs
bachp: Thanks for the feedback. I did some initial research into .netrc and thus opened https://github.com/seanmonstar/reqwest/issues/2455.

I will give it another go as I also think it is a better solution.

TitanNano:
I tested this PR and unfortunately, I cannot make it work with our GitLab instance. Our instance only permits access token authentication (personal access token, project access token, oauth access token) or SSH. On the other hand, none of the Git credentials helpers I tested support the `authtype=bearer` credentials. `git-credential-osxkeychain` only supports username and password, `git-credential-oauth` uses `username=oauth2` and `password=[access_token]`, and `git-credential-store` also only supports username + password. This has the effect that `cargo-vet` sends any of these credentials as basic-auth while the GitLab API expects a `Authorization: Bearer [token]` header.

GitLab itself circumvents this issue by permitting any username in the HTTP authentication when fetching git repos, but the API is not that flexible.

TitanNano:
Another issue is that now every URL requires credentials. Even `raw.githubusercontent.com` and `index.crates.io`....

src/credentials.rs
```diff
@@ -0,0 +1,274 @@
...
+
+/// Execute git credential fill for the given input
+///
+/// This function is separated to allow for easier testing by mocking the git call.
+async fn execute_git_credential_fill(
+    credential_input: &str,
+) -> Result<Option<String>, Box<dyn std::error::Error>> {
+    // Execute git credential fill
+    let mut child = tokio::process::Command::new("git")
+        .args(&["credential", "fill"])
```
TitanNano, line 52: This will prevent `git credential` from prompting for credentials of unknown domains:

```suggestion
        .args(&["credential", "fill"])
        .env("GIT_TERMINAL_PROMPT", "0")
```

src/credentials.rs
```diff
@@ -0,0 +1,274 @@
...
+impl GitCredential {
+    /// Apply the credential to a reqwest RequestBuilder
+    pub fn apply_to_request(
+        self,
+        request_builder: reqwest::RequestBuilder,
+    ) -> reqwest::RequestBuilder {
+        match self {
+            GitCredential::UsernamePassword { username, password } => {
+                request_builder.basic_auth(username, Some(password))
+            }
```
TitanNano, line 35: Support the usernames `oauth2` and `access_token` as bearer auth: 

```suggestion
            GitCredential::UsernamePassword { username, password }
                if username == "oauth2" || username == "access_token" =>
            {
                request_builder.header("Authorization", format!("Bearer {password}"))
            }
            GitCredential::UsernamePassword { username, password } => {
                request_builder.basic_auth(username, Some(password))
            }
```

PR #645, opened 2024-12-20
Upgrade enough dependencies to upgrade to ring 0.17

glandium:
while limiting the amount of new and/or duplicate crates. ring 0.17 contains a build fix for arm64 windows.


woodruffw:
Sorry for the necropost, but I'm curious if there's anything I or others can do to help push this along -- ring 0.16 is described as "completely unmaintained" and is also hard/impossible to build against a Windows aarch64 target, making it difficult to redistribute `cargo-vet` for that platform.

Ref: https://github.com/briansmith/ring/discussions/2450

PR #648, opened 2025-01-09
 Don't prompt to renew expiring wildcard audits for inactive crates

### LINKED ISSUE #647: Don't warn about expired/expiring wildcard audits if the crate hasn't been updated in a while
Right now cargo-vet on mozilla-central produces the following:

```
Vetting Succeeded (385 fully audited, 75 partially audited, 56 exempted)
 WARN Your audit set contains wildcard audits for bhttp, cexpr, ohttp, dogear, and 17 others which have expired.
 WARN Your audit set contains wildcard audits for qcms, glean, mozdevice, and 6 others which expire within the next six weeks.
 WARN   Consider running `cargo vet renew --expiring` or adding `renew = false` to the wildcard entries in audits.toml.
```

One thing I noticed while trying to refresh them is that many of the crates haven't seen updates in quite some time. Conceptually, the point of wildcard audits is to ensure that future versions of a crate developed by the same person can be automatically covered. However, for crates that are not under active development, a wildcard audit becomes an ongoing maintenance hassle due to the need to renew. You don't quite want to mark them as `renew = false` (since that's something you'd do if the crate transferred ownership or the owner left the project), but it's also annoying to keep renewing.

A simple heuristic would be to skip the warning if the crate hasn't been updated in the last, say, 4 months. I suspect this would be trivial to do. @mystor WDYT?

mystor:
This silences the wildcard audit renewal warning for crates which have not been updated in at least 4 months, and haven't been updated since the wildcard audit has expired.

Fixes #647

### DIFF
--- a/src/git_tool.rs
+++ b/src/git_tool.rs
@@ -136,7 +136,7 @@ pub struct Editor<'a> {
     run_editor: Box<dyn FnOnce(&Path) -> io::Result<bool> + 'a>,
 }
 
-impl<'a> Editor<'a> {
+impl Editor<'_> {
     /// Create a new editor for a temporary file.
     pub fn new(name: &str) -> io::Result<Self> {
         let tempfile = tempfile::Builder::new()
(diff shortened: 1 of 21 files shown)


afranchuk (approved):
This is great!

One minor nitpick is that I think the tests could use the `WILDCARD_AUDIT_INACTIVE_CRATE_DURATION` so that the tests are done relative to that value (if it changes, the tests won't start failing). Likewise some of the tests have comments which explicitly mention "16 weeks", but to avoid them getting out of sync it could just say "the expiry duration" or something like that.

mystor:
I'll mention that time in comments, but I think I prefer having the test independently verify the time spans rather than deriving the spans directly from the value used in main.rs.

PR #649, opened 2025-01-20
Don't automatically renew wildcard audits for inactive crates

mystor:
This is a follow-up from #648. After that change the `cargo vet renew --expiring` command still would renew audits for inactive crates. This changes that behaviour to also require an `--include-inactive` flag to renew for inactive crates.

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -750,6 +750,11 @@ pub struct RenewArgs {
     #[clap(long, action, conflicts_with("crate-name"))]
     pub expiring: bool,
 
+    /// Renew wildcard audits for inactive crates which have not been updated
+    /// in 4 months.
+    #[clap(long, action, requires("expiring"))]
+    pub include_inactive: bool,
+
     /// The name of a crate to renew.
     #[clap(value_name("CRATE"), action, required_unless_present("expiring"))]
     pub crate_name: Option<String>,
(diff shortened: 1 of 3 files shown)


src/cli.rs
```diff
@@ -750,6 +750,11 @@ pub struct RenewArgs {
     #[clap(long, action, conflicts_with("crate-name"))]
     pub expiring: bool,
 
+    /// Renew wildcard audits for inactive crates which have not been updated
+    /// in 4 months.
+    #[clap(long, action, requires("expiring"))]
+    pub include_inactive: bool,
```
afranchuk, line 756: Maybe we want this to optionally indicate the inactive period (to override the default 4 months)?

afranchuk (approved):
I have a very minor feature suggestion (that likely isn't particularly useful to anyone), otherwise this looks good!

src/cli.rs
mystor: I imagine we'd would want that to be a separate flag to specify the inactive period to ignore, as this flag currently completely eliminates the inactive period (which I guess would be ~equivalent to setting it to ∞ years). I'm not sure how much value something like that would have, so I'm a bit inclined to not add it unless we decide it's necessary.

If we did have a way to configure the inactive period, I imagine it would be more useful as a field in `config.toml` or similar, so that it would also impact the warnings from `cargo vet`.

PR #653, opened 2025-03-06
Advertise Google's ub-risk-N audit criteria as an example of custom criteria

riking:
The [ub-risk-N criteria set](https://github.com/google/rust-crate-audits/blob/main/auditing_standards.md#ub-risk-0) seems to be a well-designed set of audit criteria that seems to have worked well for over a year, and aside from zcash's `crypto-reviewed` and `license-reviewed` is the only set of non-default criteria currently available in the ecosystem registry.toml.

I didn't include the `does-not-implement-crypto` + `crypto-safe` criteria set because it seems fairly obvious, and I didn't include the `rule-of-two-safe-to-deploy` criteria because the audit registry doesn't actually include any audits under that criteria, so it's not actually in use in any visible manner.

I have *not* consulted anyone from Google about adding this, but it does seem like a very high quality document that should get additional attention.

In particular, this scheme adds an obvious way to record "qualified" and "adverse" audits with ub-risk-3 and "ub-risk-4 without safe-to-run" that is more clear than `criteria = []` which may look like an auditor mistake.

### DIFF
--- a/book/src/audit-criteria.md
+++ b/book/src/audit-criteria.md
@@ -51,3 +51,20 @@ criteria:
   for A, B, and C allows the two projects to partially share work.
 
 You can define and use as many separate sets of criteria as you like.
+
+### Example criteria set: Google's `ub-risk-N`
+
+Google's rust crate audits define a set of 7 audit criteria that form an
+implication chain: `ub-risk-0` through `ub-risk-4`, along with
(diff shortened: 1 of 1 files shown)


mystor (approved):
This seems reasonable enough to include as an example. Thanks.

PR #658, opened 2025-06-03
Update some dependencies to more recent versions

mystor:
This patch updates some of the more significant dependencies of cargo-vet to more recent versions of the corresponding crate.

The following are the notable updates:

* `toml` and `toml_edit` have adjusted their API and formatting of TOML outputs (which needed to be accounted for to avoid disruptive formatting changes). In addition, the error format changed to include more details, which lead to toml parse error changes.
* `crates-index` was updated to the latest version.
* The dependency on `winapi` was replaced with a dependency on `windows-sys` for file locking.
* Argument parsing was updated to `clap 4`. This was a substantial API and help formatting change, which necessitated a change to markdown-help generation.

In addition, a number of trusted entries and audits were added to allow cargo vet to pass once again. A few crates for unsupported targets (specifically redox) were explicitly exempted.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -56,7 +56,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions-rs/toolchain@v1
         with:
-          toolchain: 1.70.0
+          toolchain: 1.82.0
           override: true
       - name: Run cargo check
         run: |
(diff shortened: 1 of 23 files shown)


afranchuk (approved):
I like the code clean-up that fell out of this!

PR #659, opened 2025-06-03
Add a subcommand to explain how a crate has been audited

### LINKED ISSUE #654: Add a command to answer the question: "has anyone I trust reviewed $crate"
Story time: I was looking at https://phabricator.services.mozilla.com/D242217 and wondered if there was an update to the serde_cbor crate that would bump its dependency on half, at which point I figured out it's unmaintained. Which brought me to wonder about alternatives, and the serde_cbor repo itself lists a couple.

The way it is right now, I'd need to effectively add an actual dependency for `cargo vet` to tell me whether there are audits available. A query mode that doesn't require that would be useful is such situations. Ideally, it would also go through the queried crate's dependencies as well.

mystor:
Fixes #654


afranchuk (changes requested):
I like this feature! I have some minor questions/comments.

PR #662, opened 2025-08-11
Use audit versions to check audit-as-crates-io policies.

### LINKED ISSUE #661: Locked runs (CI) will not catch a crate being switched to a locally patched version
In https://phabricator.services.mozilla.com/D259200, we patched the vendored copy of `zip` to be specifically the local version, rather than a vendored copy of the one from crates.io. This should have required the author to add an entry to `config.toml`, like:

```toml
[policy.zip]
audit-as-crates-io = true
```

As we now have a local dependency in the tree which has a matching version number to a crate published on crates.io, and we still want to require cargo-vet audits for the crate. We would have enforced this had the patch author run `cargo vet` locally, but we don't check the registry when running on CI (to avoid new errors appearing due to new crates being pushed to the registry).

In specific cases like this it would be nice to detect, even when on CI, if we have an existing audit for the exact version of a crate that we probably want to have an `audit-as-crates-io` entry for that crate, so that we can catch this specific issue on CI, rather than when another developer tries to run `cargo vet` on the new tree.

afranchuk:
Closes #661.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -1951,7 +1951,7 @@ async fn fix_audit_as(
                 cfg.metadata
                     .packages
                     .iter()
-                    .filter(|&p| (*p.name == error.package))
+                    .filter(|&p| *p.name == error.package)
                     .map(|p| p.vet_version())
                     .collect()
             };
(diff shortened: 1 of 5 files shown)


src/main.rs
```diff
@@ -3233,6 +3233,37 @@ async fn check_audit_as_crates_io(
                 errors: shouldnt_be_audit_as,
             }));
         }
+    } else {
+        // When we don't have a network, we can still infer a few things about crates.io state.
```
mystor, line 3237: Can we actually do this particular branch all of the time, even if we have network? Helps avoid situations where errors only appear on CI if this somehow falls out-of-sync with the above checks.

I think this would look like removing the big check above, and instead doing the network check in the iterator loop.

Roughly I think the idea would look like:

```rust
// Do we have local audits or exemptions which name this crate, indicating that
// at one point we cared about auditing this crate due to it being on crates.io?
let has_local_audits = store.audits.audits.has_key(&package.name) ||
    store.audits.wildcard_audits.has_key(&package.name) ||
    store.audits.trusted.has_key(&package.name) ||
    store.config.exemptions.has_key(&package.name);

// To do some validation when no network is available, we also assume the crate
// is on crates.io if there are any local audits for the crate name (to avoid the crate
// being missed when e.g. applying local patches).
let crate_is_on_crates_io = matches_crates_io_package || has_local_audits;

if crate_is_on_crates_io && audit_policy.is_none() {
  // We found a package that has similar metadata to one with the same name
  // on crates.io, or we have local audits for this crate name: having no policy is an error.
  return Some((CheckAction::NeedAuditAs, package));
}

if network.is_some() && !crate_is_on_crates_io && audit_policy == Some(true) {
  return Some((CheckAction::ShouldntBeAuditAs, package));
}
```

src/main.rs
```diff
@@ -3233,6 +3233,37 @@ async fn check_audit_as_crates_io(
...
+        let mut needs_audit_as_entry = vec![];
+        'for_package: for package in first_party_packages
+            .iter()
+            .filter(|p| p.git_rev().is_none())
+        {
+            for audits_file in std::iter::once(&store.audits).chain(store.imports.audits.values()) {
+                let Some(entries) = audits_file.audits.get(package.name.as_str()) else {
```
mystor, line 3244: I think you're being too precise here, we probably want to require audit_as if the crate name was mentioned at all in audits/wildcard audits/exemptions/trusted, as that indicates that at some point in the past we cared about auditing this crate.

mystor (changes requested):

PR #673, opened 2025-11-05
Fix the regeneration of audit-as-crates-io entries.

afranchuk:
There was a case where regeneration hit an assertion: if there existed a versioned crate policy and a local crate of the same name, the regeneration would try to add a non-versioned policy for the local crate and fail.

This changes the behavior to add a versioned policy for the local crate in this case.

### DIFF
--- a/tests/test-cli.rs
+++ b/tests/test-cli.rs
@@ -44,17 +44,22 @@ fn format_diff_outputs(output: &Output) -> String {
         .lines()
         .filter(|line| !line.starts_with("diff --git"))
         .map(|line| {
+            fn simple_path(path: &str) -> &str {
+                // We trim `"` because the git diff paths may be quoted and escaped as described in
+                // `core.quotePath` if the parent directories have unusual characters. We don't
+                // need to worry about handling the escape sequences: the filenames that we use
+                // have no such characters, and we just use the filename.
+                Path::new(path.trim_matches('"'))
+                    .file_name()
+                    .unwrap()
+                    .to_str()
+                    .unwrap()
+            }
             if let Some(path) = line.strip_prefix("--- ") {
-                return format!(
-                    "--- a/{}",
-                    Path::new(path).file_name().unwrap().to_str().unwrap()
-                );
+                return format!("--- a/{}", simple_path(path));
             }
             if let Some(path) = line.strip_prefix("+++ ") {
-                return format!(
-                    "+++ b/{}",
-                    Path::new(path).file_name().unwrap().to_str().unwrap()
-                );
+                return format!("+++ b/{}", simple_path(path));
             }
             line.to_owned()
         })


src/format.rs
```diff
@@ -698,7 +699,16 @@ impl Policy {
 
         match (pkg_policy, version) {
```
mystor: So the `get_mut_or_default` method has only 1 caller in the codebase, which appears to be in `fix_audit_as`.

I'm having a hard time following the argument behind the unwrap in that caller.

1. The `PackageError` type has an optional `version` member, but it looks like every single place it's initialized as `Some(package.vet_version())` so perhaps we can drop the `Option<...>`?
2. Over-versioning seems like not the end of the world, so I think I bias in that direction
3. It seems like it might make sense to always ensure that all versions are represented
4. It's not 100% clear to me why we need to have an entry for every single version 

I wonder whether it could instead work something like this, but maybe I'm missing some part of why this would be a problem:

```rust
    /// Get the mutable policy entry for the given crate, creating a default if none exists.
    ///
    /// Unlike `get_mut`, if `require_versions` is provided, this guarantees
    /// that the policy is represented as versioned. This will never convert a
    /// versioned policy entry to unversioned.
    ///
    /// When converting an unversioned policy entry to versioned, the existing
    /// unversioned entry will be used for all new versioned entries which are
    /// being inserted.
    pub fn get_mut_or_default(
        &mut self,
        name: PackageStr<'_>,
        version: &VetVersion,
        require_versions: Option<Vec<VetVersion>>,
    ) -> &mut PolicyEntry {
        // Ensure a policy entry exists for this package, inserting a default
        // unversioned entry if it does. This will be converted to a versioned
        // entry below if needed.
        let pkg_policy = self
            .package
            .entry(name.to_owned())
            .or_insert_with(|| PackagePolicyEntry::Unversioned(Default::default()));

        // If we need a versioned entry, make sure that the entry is configured
        // as versioned, and insert them.
        if let Some(require_versions) = require_versions {
            // By default we insert an empty policy, but if there was an
            // existing unversioned entry, we'll use that as the policy for all
            // entries.
            let mut default = PolicyEntry::default();
            let versions_map = loop {
                match pkg_policy {
                    PackagePolicyEntry::Unversioned(entry) => {
                        default = mem::take(entry);
                        *pkg_policy = PackagePolicyEntry::Versioned {
                            version: Default::default(),
                        }
                    }
                    PackagePolicyEntry::Versioned { version } => break version,
                }
            };
            for version in require_versions {
                versions_map
                    .entry(version)
                    .or_insert_with(|| default.clone());
            }
        }

        match pkg_policy {
            PackagePolicyEntry::Unversioned(e) => e,
            PackagePolicyEntry::Versioned {
                version: versions_map,
            } => versions_map.entry(version.clone()).or_default(),
        }
    }
```

mystor (changes requested):

mystor:
As we discussed separately the windows failures appear to be due to a change in the output of `git diff` on Windows, combined with https://github.com/mozilla/cargo-vet/blob/432fdc99ac20bd988ab36a64791ae35560f0f6b2/tests/test-cli.rs#L47-L59 not shell-unescaping the paths, leading to an extra `"` at the end. We can perhaps do better here.

src/format.rs
afranchuk: Oh, I think I misunderstood your point. I thought you were suggesting we remove `Option` from the `PackageError::version` field.

PR #678, opened 2026-01-14
Update cargo-dist to 0.30.3.

afranchuk:
cargo-dist 0.30.1 updated the default macos runners to macos-15, as macos-13 was deprecated at the time (and has now been removed entirely, causing distribution to fail without this update).


mystor (approved):
Thanks! Is this something which we can use to re-run the publish actions for 0.10.1 and 0.10.2, or will we need to do something else for that?

afranchuk:
> Thanks! Is this something which we can use to re-run the publish actions for 0.10.1 and 0.10.2, or will we need to do something else for that?

I believe if you re-tag them (which might involve pushing a tag delete followed by pushing a new tag, since otherwise there would be no difference) it should re-run the publish?

eems-jwimmer:
Any follow-up here? I'm not sure if Mozilla or this repo in particular enforces immutable tags, but if not, you could merge and retag. Otherwise, you'll likely need to bump a patch version.

PR #685, opened 2026-04-18
Update links in importing-audits.md to mozilla github organization

danjl1100:
Hello,

When learning about `cargo-vet`, I was surprised that the link to the registry (list of organizations to possibly consider trusting) pointed to what appeared to be another user's `cargo-vet` repo.

After a cursory glance, I see that the github URL does redirect to the mozilla organization, so I imagine this is just a small update missed when the repo was transferred to the mozilla organiztion.  It's just the raw.githubusercontent URL that doesn't auto-redirect.

I usually like to open an issue first to discuss, but this one is so small I'm just sending the PR. (these were the only URLs I found mentioning bholley, which might have been missed in the migration)
Feel free to close if not needed, or even squash into other doc change that are queued :slightly_smiling_face:

### DIFF
--- a/book/src/importing-audits.md
+++ b/book/src/importing-audits.md
@@ -33,9 +33,9 @@ desired to locally-defined criteria.
 
 To ease discovery, `cargo vet` maintains a central registry of the audit sets
 published by well-known organizations. This information is stored in the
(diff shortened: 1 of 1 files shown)


bholley (approved):
LGTM thanks!

